PR #2, opened 2021-07-02
chore(build): add gh actions trigger to codebuild

sfoster1:
To avoid having to set up aws webhooks, we can try triggering the codebuild with [this aws-maintained action](https://github.com/aws-actions/aws-codebuild-run-build) and posting the result in slack.


sotaoverride (approved):
LGTM!

PR #3, opened 2021-07-08
New config files

sotaoverride:
Updating config files for new recipes added.

### DIFF
--- a/build/conf/local.conf
+++ b/build/conf/local.conf
@@ -1,4 +1,8 @@
-IMAGE_INSTALL_append = " python3-modules python3-robot-shared-data python3-robot-api python3-update-server python3-notify-server"
+IMAGE_INSTALL_append = " open python3-robot-shared-data update-server python3-notify-server python3-robot-api robotserversystemd python3-uvicorn"
+DISTRO_FEATURES_append = " systemd"
+DISTRO_FEATURES_BCKFILL_CONSIDERED += "sysvinit"
+VIRTUAL-RUNTIME_init_manager = "systemd"
+VIRTUAL-RUNTIME_initscriots = "systemd-compat-units"
 
 ACCEPT_FSL_EULA = "1"
 #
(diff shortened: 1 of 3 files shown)


build/conf/local.conf
```diff
@@ -1,4 +1,8 @@
-IMAGE_INSTALL_append = " python3-modules python3-robot-shared-data python3-robot-api python3-update-server python3-notify-server"
+IMAGE_INSTALL_append = " python3-robot-server python3-robot-shared-data python3-update-server python3-notify-server python3-robot-api robotserversystemd python3-uvicorn"
+DISTRO_FEATURES_append = " systemd"
+DISTRO_FEATURES_BCKFILL_CONSIDERED += "sysvinit"
+VIRTUAL-RUNTIME_init_manager = "systemd"
+VIRTUAL-RUNTIME_initscriots = "systemd-compat-units"
```
sfoster1, line 5: typo?
```suggestion
VIRTUAL-RUNTIME_initscripts = "systemd-compat-units"
```

build/conf/local.conf
```diff
@@ -1,4 +1,8 @@
-IMAGE_INSTALL_append = " python3-modules python3-robot-shared-data python3-robot-api python3-update-server python3-notify-server"
+IMAGE_INSTALL_append = " python3-robot-server python3-robot-shared-data python3-update-server python3-notify-server python3-robot-api robotserversystemd python3-uvicorn"
+DISTRO_FEATURES_append = " systemd"
+DISTRO_FEATURES_BCKFILL_CONSIDERED += "sysvinit"
```
sfoster1, line 3: typo?
```suggestion
DISTRO_FEATURES_BACKFILL_CONSIDERED += "sysvinit"
```

sfoster1: Also do we actually want backfill here? There's a lot of stuff that entirely relies on systemd.

sotaoverride: Think this is to keep sysvinit from getting backfilled.. if a distro feature isn't set, backfilled ones get added automatically. this is because we added new features and made them default.  if we don't want it, we can add it to DISTRO_FEATURES_BACKFILL_CONSIDERED to say that we know it would  be backfilled, but we don't want it.

sfoster1: oh, i see, the semantics are "i personally considered (and rejected) this, bitbake you don't have to"

sfoster1 (approved):
lgtm!

PR #5, opened 2021-07-12
fix(build): Correct cf output url

sfoster1:
The build id output from the codebuild action encoded the build project into the name, which we don't want, so remove it with a bashism. Also fix some other stuff about the url and make it (slightly) more readable.


sotaoverride (approved):
LGTM, that url hopefully works now.

PR #6, opened 2021-07-15
chore: add netboot docker server

sfoster1:
It will be a lot easier to do development on the verdin without having
to constantly bounce around between recovery mode and flashing the
internal mmc. This adds a docker container and instructions for host
system standup (it is not that isolated) that can connect directly to a
verdin via an ethernet interface, and provide a networked bootloader
chain given a u-boot config on the verdin that accepts it.

The docker container, when bound to the appropriate iface, provides a
dhcp server, which also provides a bootp kernel download; and then an
nfs filesystem server for the root filesystem.

When this all works, with the makefile provided, it should allow devs to
carve out a directory for a filesystem unzipped from an openembedded
build, and boot a verdin off their local machine, with the ability to
much more easily change files "on" the device.

Also this does require some host machine setup, and the host machine
really has to be linux.


dev/nfs/Makefile
```diff
@@ -0,0 +1,211 @@
...
+  $(WORKDIR)/config/dhcpd.conf\
+  $(WORKDIR)/config/exports\
+  $(WORKDIR)/config/tftpd-hpa
+
+.PHONY: docker-build
+docker-build:\
+  $(WORKDIR)/config/exports \
+  $(WORKDIR)/config/dhcpd.conf \
+  $(WORKDIR)/config/tftpd-hpa \
+  $(WORKDIR)/provide/boot $(WORKDIR)/provide/root \
```
sotaoverride, line 95: make seems to be ignoring these prereqs, i was seeing this
`bash: line 0: cd: work/provide/boot: No such file or directory`

sotaoverride: guess it was just paths like these
cd $(WORKDIR)/provide/boot
testing it out with absolute paths instread

dev/nfs/Makefile
```diff
@@ -0,0 +1,212 @@
...
+	-docker stop $(CONTAINER_NAME)
+	-docker rm $(CONTAINER_NAME)
+
+
+# If an image was provided, copy it in; otherwise roll with what's present
+ifdef IMAGE
+.PHONY: set-image
+set-image: $(WORKDIR)
+	$(RM) -r $(WORKDIR)/unzip
+	mkdir -p $(WORKDIR)/unzip
```
sotaoverride, line 171: do we need `mkdir -p $(WORKDIR)/provide` under the recipe portion of this target?

PR #7, opened 2021-07-27
fix(start.sh bblayers.conf): copy wic image at end of build. make met…

sotaoverride:
oe-core changes
Make meta-opentrons priority the highest. There are some machine definitions with the same names in meta-opentrons and other meta-layers coming in from the bsp. While we're still hanging on to some of the bsp provided classes, meta-layers, etc, we can make the priority for meta-opentrons the highest (by just moving it up in the bblayers.conf list). 

Image with A/B partitions is a wic image, copying that over from the container after build completes.

### DIFF
--- a/build/conf/bblayers.conf
+++ b/build/conf/bblayers.conf
@@ -6,6 +6,7 @@ BBPATH = "${TOPDIR}"
 BBFILES ?= ""
 
 BBLAYERS ?= " \
+  ${TOPDIR}/../layers/meta-opentrons \
   ${TOPDIR}/../layers/meta-toradex-nxp \
   ${TOPDIR}/../layers/meta-freescale \
   ${TOPDIR}/../layers/meta-freescale-3rdparty \
@@ -26,5 +27,4 @@ BBLAYERS ?= " \
   ${TOPDIR}/../layers/meta-toradex-distro \
   ${TOPDIR}/../layers/meta-yocto/meta-poky \
   ${TOPDIR}/../layers/openembedded-core/meta \
-  ${TOPDIR}/../layers/meta-opentrons \
   "
--- a/start.sh
+++ b/start.sh
@@ -16,8 +16,11 @@ patch ./layers/meta-toradex-nxp/recipes-kernel/linux/linux-toradex_5.4-2.3.x.bb 
 export BITBAKEDIR=${THISDIR}/tools/bitbake
 . layers/openembedded-core/oe-init-build-env ${THISDIR}/build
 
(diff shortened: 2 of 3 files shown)


build/conf/bblayers.conf
```diff
@@ -6,6 +6,7 @@ BBPATH = "${TOPDIR}"
 BBFILES ?= ""
 
 BBLAYERS ?= " \
+  ${TOPDIR}/../layers/meta-opentrons \
```
sfoster1, line 9: Does this mess with recipe/dependency overrides?

start.sh
```diff
@@ -16,8 +16,11 @@ patch ./layers/meta-toradex-nxp/recipes-kernel/linux/linux-toradex_5.4-2.3.x.bb
 export BITBAKEDIR=${THISDIR}/tools/bitbake
 . layers/openembedded-core/oe-init-build-env ${THISDIR}/build
 
-BB_NUMBER_THREADS=$((`nproc`-1)) bitbake tdx-reference-minimal-image
+BB_NUMBER_THREADS=$((`nproc`-1)) bitbake opentrons-ot3-image 
 
 cd ${THISDIR}
 mkdir -p build/deploy/opentrons
-cp $(find build/deploy/images/verdin-imx8mm/ | grep Reference-Minimal-Image-Tezi | head -n 1) build/deploy/opentrons/opentrons-image.tar
+cp $(find build/deploy/images/verdin-imx8mm/ | grep opentrons-ot3-image-Tezi | head -n 1) build/deploy/opentrons/opentrons-image.tar
```
sfoster1, line 23: do we need to build two images or does the one image recipe make both the tezi image and the wics image?

build/conf/bblayers.conf
sotaoverride: this just makes sure if there a recipe with the same name, meta-opentron's version of the recipe gets used.

start.sh
sotaoverride: one recipe would make both the images

sfoster1 (changes requested):
Can you edit the workflow to also post the wics image in the slack message by adding an output here: https://github.com/Opentrons/oe-core/blob/42ac853f269d59801c0db66aa3d4cdc387ec890a/.github/workflows/build-ot3.yml#L32 and then putting it in the formatted message here: https://github.com/Opentrons/oe-core/blob/42ac853f269d59801c0db66aa3d4cdc387ec890a/.github/workflows/build-ot3.yml#L36

sfoster1:
Looks good tho, downloaded the image and all

sfoster1 (approved):
Looks good to me!

sotaoverride:
ci didn't like this branch name.

PR #8, opened 2021-07-27
fix(start.sh bblayers.conf): copy wic image at end of build. make met…

sotaoverride:
oe-core ab partition changes
Make meta-opentrons priority the highest. There are some machine definitions with the same names in meta-opentrons and other meta-layers coming in from the bsp. While we're still hanging on to some of the bsp provided classes, meta-layers, etc, we can make the priority for meta-opentrons the highest (by just moving it up in the bblayers.conf list).

Image with A/B partitions is a wic image, copying that over from the container after build completes.


sfoster1 (approved):
Wait this is the right one. Looks good to me!

PR #9, opened 2021-07-30
chore: gitignore

sfoster1:
We really need a gitignore oh boy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,9 @@
+.DS_Store
+.idea
+*.log
+tmp/
+
+build/
+-build/conf
+downloads/
+sstate-cache/


.gitignore
```diff
@@ -0,0 +1,9 @@
+.DS_Store
+.idea
+*.log
+tmp/
+
+build/
+-build/conf
```
sotaoverride, line 7: not sure what the - in -build/conf means, but I think we need to hold on to build/conf (it has bblayers and local.conf in it)?

sotaoverride (commented):
build/conf, shouldn't we be holding on to it?

.gitignore
sotaoverride: or nevermind, think the -build/conf means we are holding on to the conf dir..

sotaoverride (approved):
lgtm

.gitignore
```diff
@@ -0,0 +1,9 @@
+.DS_Store
+.idea
+*.log
+tmp/
+
+build/
```
sotaoverride, line 6: should we add something like -build/deploy, deploy shouldnt be ignored

PR #10, opened 2021-07-30
chore(build): slight build speedups, more capability

sfoster1:
You can now specify arguments to `ot3Update.sh` and have them forwarded
to bitbake. In addition, we now under the hood build the docker image
with a tempdir context that's empty except for `start.sh`, so it's a lot faster.


sotaoverride (approved):
lgtm

PR #11, opened 2021-08-02
chore(build): reenable caching on codebuild

sfoster1:
This should speed things up some hopefully. We can get more aggressive
by actually caching build results, but doing the state and the downloads
should help.


sfoster1:
This won't work until we change to S3 caching

sotaoverride (approved):
lgtm

PR #17, opened 2021-10-21
core: re-add download caching

sfoster1:
The thing we were missing  with previous caching effort was saving the [sstate cache](https://wiki.yoctoproject.org/wiki/Enable_sstate_cache) along with the downloads folder. By caching both sstate and downloads, we bring the builds down to about an hour. We can also save buildstats for later perusal and put them in the artifact store.


sotaoverride (approved):
lgtm

PR #19, opened 2022-07-07
layers: meta-freescale: use dunfell-fix-gtk branch

sfoster1:
This is the commit we need to use and the one that matches .gitmodules.
Without this, gdk doesn't build its x11 support, which is all our
version of chromium from electron 13 can use, and doesn't work.


sfoster1:
This works on a device (though the app moves on to some js stack errors, it no longer crashes because gtk doesn't have its x11 backend).

PR #20, opened 2022-08-26
RET-1205 Remove recipe source files are they are built to save diskspace

vegano1:
[RET-1205](https://opentrons.atlassian.net/browse/RET-1205)

CI build [oe-core-run:8039613853](https://github.com/Opentrons/oe-core/runs/8039613853) is failing because we run out of diskspace. So add rm_recipe to the local.conf file which will remove recipe source files after the recipe is built.

### DIFF
--- a/build/conf/local.conf
+++ b/build/conf/local.conf
@@ -41,9 +41,6 @@ MACHINE ?= "verdin-imx8mm"
 #MACHINE ?= "qemux86"
 #MACHINE ?= "qemux86-64"
 
-INHERIT += "toradex-mirrors rm_work"
-# exlude opentrons-ot3-image
-RM_WORK_EXCLUDE += "opentrons-ot3-image"
 
 MACHINEOVERRIDES =. "use-head-next:"
 #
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Looks good to me!

build/conf/local.conf
```diff
@@ -265,7 +265,7 @@ CONF_VERSION = "1"
 # Delete the the source/object/binary files once a package is built to preserve disk space
```
SyntaxColoring, line 265: Should `rm_work` be added directly under this comment instead?

It looks like that's how things were back in https://github.com/Opentrons/oe-core/commit/eaa5c3d595a695127c2a7fb71ab4e1570a9dcad3#.

sfoster1: Oh yeah we should move the include or the comment good catch

sfoster1 (changes requested):
Hold on we've got it in two places now. The inherit is on line 44 and 271. Let's get it in just one place - maybe keep the one on 271 just being the toradex mirrors and have a separate block for just the rm_work stuff with the comment?

build/conf/local.conf
vegano1: good catch, moved it down.

sfoster1 (approved):
Looks good, but please merge with a PR title that looks like `build/conf/local.conf: remove recipe source files as they are built...`

PR #25, opened 2022-10-25
recipes-image/opentrons-image-ot3: fix opentrons version file check + invalid import

vegano1:
- make sure we check all version files in the list and not just one variable
- fix JSONDecodeError import error


sfoster1 (approved):
Looks good to me if the build completes!

PR #27, opened 2022-11-06
recipes-images/opentrons-ot3-image: Add versioning support for update-server + build optimizations

vegano1:
**Description**
Changes required to support update-server openembedded capabilities as well as build optimizations

**Closes**
[RCORE-306](https://opentrons.atlassian.net/browse/RCORE-306)
[RCORE-320](https://opentrons.atlassian.net/browse/RCORE-320)

**Changes**
- Added robot_type, build_type, openembedded_version, openembedded_sha, openembedded_branch to VERSION.json to be used by the update-server for update validation.
- Added "autoinstall" flag to Tezi Image tar, so the update automatically starts in Easy Installer
- Added "reboot" to wrapup-ot3.sh script so the system automatically reboots after the Tezi image is installed
- Removed systemfs manual generation and started using the generated rootfs, this decreased build time by ~30%
- Trim userfs.ext4 by creating a smaller sparse file that's only 3% larger than the userfs dir
- Fix issue with generated systemfs.sha256 file which included the name of the file instead of just the hash


sfoster1 (changes requested):
I think this makes us have the contents of `/home` and `/var` in _both_ systemfs and userfs, since we used to use rsync with `--delete-excluded` and we don't anymore. we need to `rm` those directories instead. Otherwise, looks good!

sfoster1 (approved):
Looks great to me! Great work!

PR #31, opened 2022-11-10
recipes-robot: A few changes around the stack to address On-Device-Display issues

vegano1:
[RCORE-343](https://opentrons.atlassian.net/browse/RCORE-343)

**Overview**
Addressing some issues discovered after updating the rootfs from the opentrons desktop app

**Changes**
- Change ROBOT_MODEL to ROBOT_TYPE so we put the correct value when creating the VERSION.json file
- Added DSI-1 output to weston.ini to rotate the On-Device-Display -180 degrees so it shows up with the correct orientation 
- Added IsOnDevice=1 flag when starting opentrons-robot-app.sh script so the app knows its running on the ODD
- Added overlay to drive - DISPL_RST_CTRL_SOM - SODIMM_44 high on startup so the LCD turns on

**Todo**
- [x] validate the overlay change on the actual ODD hardware

### DIFF
--- a/layers/meta-opentrons/recipes-graphics/wayland/files/imx/weston.ini
+++ b/layers/meta-opentrons/recipes-graphics/wayland/files/imx/weston.ini
@@ -8,10 +8,9 @@ idle-time=0
 #[shell]
 #size=1920x1080
 
-#[output]
-#name=HDMI-A-1
-#mode=1920x1080@60
-#transform=90
+[output]
+name=DSI-1
+transform=rotate-180
 
 #[output]
 #name=HDMI-A-2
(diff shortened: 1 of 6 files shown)


layers/meta-opentrons/recipes-graphics/wayland/files/imx/weston.ini
```diff
@@ -8,10 +8,10 @@ idle-time=0
 #[shell]
 #size=1920x1080
 
-#[output]
-#name=HDMI-A-1
-#mode=1920x1080@60
-#transform=90
+[output]
+name=DSI-1
+mode=1024x600@56.1
```
sfoster1, line 13: why the lower refresh rate?

sfoster1 (approved):
Looks good, but would you mind
- making this 3 isolated commits (so we can rebase-merge and have them separate in history)
- clarifying the weston.ini modeset string which doesn't quite match the kernel modeset

layers/meta-opentrons/recipes-graphics/wayland/files/imx/weston.ini
vegano1: removed modeset

PR #32, opened 2022-11-10
build: cache using zips to fix git cache issues

sfoster1:
A series of fixes based on needed refactors and some recent build failures:
- refactor: move the aws credentialing code into an action since it was so bulky
- split caches of non-git downloads, sstate, and git downloads into separate entities
- store these entities on s3 as zips, since they contain empty directory trees that s3 will otherwise not store


vegano1:
LGTM!

vegano1 (approved):

PR #33, opened 2022-11-14
recipes-images/opentrons-ot3-image: Added persistence for wifi settings and robot name

vegano1:
[RCORE-342](https://opentrons.atlassian.net/browse/RCORE-342)

**Overview**
We were losing config settings after an update because they were stored in the systemfs which gets overwritten, so let's move the save location for the ones we can and mount the others ones with fstab where we cant.

**Changes**
- Added custom network manager conf file to change system-connections dir to /var/lib/ so they persist through updates.
- Added custom fstab for mount points instead of systemd mount units as hostname needs to be mounted early to /userfs/etc so robot names persist through updates.


layers/meta-opentrons/recipes-connectivity/networkmanager/files/NetworkManager-extra.conf
```diff
@@ -0,0 +1,2 @@
+[keyfile]
+path=/userfs/etc/NetworkManager/system-connections
```
sfoster1, line 2: I'd love to keep the existence of `/userfs` hidden from most of these configurations, it feels a lot cleaner. Instead, let's use paths from where userfs is mounted - so let's make this `/var/lib/NetworkManager/system-connections`

sfoster1: since this is a dropfile and its name doesn't matter, let's also give it a name like `system-connections-location.conf` or something

layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
```diff
@@ -50,12 +50,14 @@ MAX_SYSTEMFS_SIZE = "1536"
...
 python do_create_opentrons_manifest() {
     bb.note("Create the manifest json for for ot3-system.zip")
+    import time
     import json
     import os
 
     opentrons_manifest = {
         'robot_type': d.getVar('ROBOT_TYPE'),
         'build_type': os.getenv('OT_BUILD_TYPE', 'unknown/dev'),
+        'build_timestamp': time.time(),
```
sfoster1, line 60: let's not do this - it guarantees that subsequent builds that otherwise have the same content will not be content-equal, ruining any prospect for [build hermeticity](https://bazel.build/basics/hermeticity) (note that we almost certainly don't currently have hermetic builds, but it's a good goal to have). If we can identify the versions of everything else, we fundamentally don't really care about when the build occurred.

layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
```diff
@@ -87,20 +89,27 @@ python do_create_opentrons_manifest() {
...
 # add the rootfs version to the welcome banner
-fakeroot do_add_rootfs_version() {
+do_add_rootfs_version() {
     printf "${DISTRO_NAME} ${DISTRO_VERSION} (${DISTRO_CODENAME}) \\\n \\\l\n" > ${IMAGE_ROOTFS}/etc/issue
     printf "${DISTRO_NAME} ${DISTRO_VERSION} (${DISTRO_CODENAME}) %%h\n" > ${IMAGE_ROOTFS}/etc/issue.net
     printf "${IMAGE_NAME}\n\n" >> ${IMAGE_ROOTFS}/etc/issue
     printf "${IMAGE_NAME}\n\n" >> ${IMAGE_ROOTFS}/etc/issue.net
 
+    # add datetime as version, example 202210111213
+    date +"%Y%m%d%H%M" > ${IMAGE_ROOTFS}/etc/version
```
sfoster1, line 99: ditto here re: hermeticity, let's not put dates in the filesystem if we can help it

layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
```diff
@@ -87,20 +89,27 @@ python do_create_opentrons_manifest() {
...
     printf "opentrons" > ${IMAGE_ROOTFS}/etc/hostname
-    printf "PRETTY_HOSTNAME=opentrons\n" > "${IMAGE_ROOTFS}/etc/machine-info"
-    # TODO(ba, 2022-10-18): add proper mechanism for setting DEPLOYMENT
-    printf "DEPLOYMENT=development\n" >> "${IMAGE_ROOTFS}/etc/machine-info"
+    printf "PRETTY_HOSTNAME=opentrons\n" > ${IMAGE_ROOTFS}/etc/machine-info
+    printf "DEPLOYMENT=development\n" >> ${IMAGE_ROOTFS}/etc/machine-info
+
+    # cleanup
+    rm -rf ${IMAGE_ROOTFS}/opentrons_versions
+    rm -rf ${IMAGE_ROOTFS}/unit_tests
```
sfoster1, line 112: what's adding these?

layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
```diff
@@ -109,9 +118,19 @@ fakeroot do_create_filesystem() {
     rsync -aH --chown=root:root ${IMAGE_ROOTFS}/home ${USERFS_DIR}/
     rsync -aH --chown=root:root ${IMAGE_ROOTFS}/var ${USERFS_DIR}/
     mkdir -p ${USERFS_DIR}/data
+    mkdir -p ${USERFS_DIR}/etc
+
+    # create dir to persist network connections
+    mkdir -p ${USERFS_DIR}/etc/NetworkManager/system-connection
```
sfoster1, line 124: let's not put this here unless we're going to bind-mount it into root

sfoster1 (changes requested):
Couple things I think we should change inline:
- We don't currently have hermetic builds, but it's a good goal to have, and we should avoid introducing anything new that moves us further from this goal. So let's not put build dates or times on the filesystem
- Let's not put stuff in `/userfs/etc` that isn't bind-mounted into `/etc`

Otherwise, looks like a good step forward!

layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
vegano1: removed /unit_tests dir, it was being added by imx-vpu-hantro
as for the opentrons_versions this temporarily holds the version json files to create the VERSION.json file.

sfoster1: can we put the `unit_tests` dir in a bbappend for `imx-vpu-hantro`? want to avoid the image recipe kind of spiraling into a catchall for that stuff

sfoster1: oops looks like you did! sorry!

layers/meta-opentrons/recipes-connectivity/networkmanager/networkmanager_%.bbappend
```diff
@@ -0,0 +1,10 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+
+SRC_URI += "file://system-connections-location.conf"
+
+FILES_${PN} += "/etc/NetworkManager/conf.d/system-connections-location.conf"
+
+do_install_append() {
+	install -d ${D}/etc/NetworkManager/conf.d
+	install -m 644 ${WORKDIR}/system-connections-location.conf ${D}/etc/NetworkManager/conf.d/
```
sfoster1, line 9: may also want `install -d ${D}/var/lib/NetworkManager/system-connections`

layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
```diff
@@ -50,6 +50,7 @@ MAX_SYSTEMFS_SIZE = "1536"
 # create the opentrons ot3 manifest (VERSION.json) file
 python do_create_opentrons_manifest() {
     bb.note("Create the manifest json for for ot3-system.zip")
+    import time
```
sfoster1, line 53: don't need this anymore I believe

sfoster1 (approved):
Looks really good! A couple tiny nits, but I like it.

PR #34, opened 2022-11-16
build: use EXTERNALSRC and a custom checkout for monorepo

sfoster1:
The opentrons monorepo at https://github.com/Opentrons/opentrons contains most of the custom code that our ot3 image runs. That means that we want to change around the version of the code used pretty often. To do that with normal bitbake sourcing through `SRCURI` and `branch` would mean very frequently altering the recipe text. Instead, let's configure bitbake to look at a specific disk location for the monorepo in every recipe that uses the monorepo, and check out the appropriate revision of the monorepo ourselves in our build orchestration (and, in local builds, rely on the developer having the repo in the right place manually).

This change makes an openembedded build be an entity consisting of a source revision of oe-core and a source revision of the monorepo. Which source revision to use for each depends on which one is specified and which one is left to vary. This is accomplished with a new github action that decides appropriate revisions using the following logic, which is taken from the similar orchestration used for the monorepo and buildroot for the ot-2. 

The logic is based on some subset of used repos (right now there are 2, but in the future there will be more) which have their revisions specified. These repos are in the set `S`, such that each repo in `S` is `Si`; and the rest of the repos do not have specified revisions and are in the set `F`, such that each repo in `F` is `Fi`.

- When a build is started, identify the "authoritative ref", which shall be the first of the following repos in order that is in S: `opentrons`, `oe-core`. The authoritative ref is `A`.
- For each repo in `S`, use the specified source revision
- For each repo `Fi` in `F`
  - if `A` is the main branch of its repo (i.e. `edge` for monorepo, `main` for oe-core), use the main branch of `Fi`
  - if `A` is a branch that is not the main branch of its repo
     - and a branch of the same name exists in `Fi`, use that matching branch
     - and a branch of the same name does not exist in `Fi`, use the main branch of `Fi`
  - if `A` is a tag
    - and a tag of the same name exists in `Fi`, use that tag
    - and a tag of the same name does not exist in `Fi`, but other tags exist in `Fi`, use the latest tag in `Fi`
    - and a tag of the same name does not exist in `Fi` and there are no tags in `Fi`, use `Fi`'s main branch

The action is implemented in `.github/actions/build-refs` and is typescript that is transpiled to a checked-in (and marked as binary) js action.

## Review requests
- [ ] do you agree with the logic above
- [ ] does the action correctly implement the logic above
- [ ] style check


sfoster1:
Also I'd love some thoughts on how best to test the parts of the code that need to use octokit and how to mock that out and so on

.github/actions/build-refs/dist/index.js
```diff
@@ -0,0 +1,9898 @@
+/******/ (() => { // webpackBootstrap
```
vegano1, line 1: do we want to keep this generated code?

sfoster1: Yes. This is unfortunately how this process works - github won't autogenerate anything for you, you have to check the javascript file you want to run into the repo. Since we want to have the actual code be in typescript and transpile to javascript, we have to do that and check the code in. It's marked as a binary file which apparently github doesn't respect, which is annoying - your local git client won't show a diff here.

vegano1 (approved):

PR #35, opened 2022-11-23
recipes-bsp/u-boot: added custom u-boot script to change boot behavior

vegano1:
[RCORE-395](https://opentrons.atlassian.net/browse/RCORE-395)

# Overview
We have an A/B update mechanism in place, so we should take advantage of that and enable kernel + DTB + overlay updates by hosting them in the rootfs instead of the boot partition. This way they can be updated with a rootfs update and we don't end up like the OT2 where we can't (safely) update to a newer version when the kernel + device tree are out of date.

# Changes
- Load kernel + dtb + overlays from rootfs instead of boot partition
- switch between rootfs partitions 2-3 if boot fails


sfoster1 (approved):
Looks good to me. I like the fallback boot mode, but we may need a special kernel for it that interns modules because it probably won't reliably work without that.

PR #38, opened 2022-12-14
recipes-robot/nginx: remove nginx request size limit for /protocols endpoint

vegano1:
Valid protocols larger than 2M were failing because Nginx was configured to max request size of 2M so remove the max request size for the /protocols endpoint so we don't run into this in the future.


sfoster1 (approved):
nice!

PR #39, opened 2022-12-16
recipes-core/base-files/fstab: Change rootfs to read-only in fstab

vegano1:
**Closes:** [RCORE-178](https://opentrons.atlassian.net/browse/RCORE-178)

**Overview:**
We only want parts of the system to be read-write like /userfs, /var, etc for everything else we want read-only. 

**Changelog:**
- change root mount option to ro (read-only)
- create a symlink for dropbear host rsa key-file /etc/dropbear -> /var/lib/dropbear


sfoster1 (approved):
excellent!

PR #41, opened 2022-12-21
build: fix permission issue causing python_build_util.py to fail in Docker container.

vegano1:
# Overview:

Builds are failing because of an exception caused bu using python_build_util.py when attempting to run a git command as a subprocess. This was due to /volumes/opentrons not having the right permissions, so lets just label /volumes/opentrons as a safe.directory in the git config.

# Changes:

- set /volumes/opentrons as a safe.directory in the git config of the Dockerfile


sfoster1 (approved):
Nice, sorry about that!

PR #42, opened 2022-12-23
oe-core: Update to toradex 6.0.0 and OE kirkstone

sfoster1:
# Summary

This PR updates our openembedded stack to version [6.0.0 of the Toradex yocto BSP](https://www.toradex.com/news/bsp-layers-reference-images-yocto-project-6-0-0-quarterly) (was 5.x). Fixed issues in the BSP can be found [here](https://developer-archives.toradex.com/software/linux/release-details?status=21&module=Verdin+iMX8M+Mini).

# Major Updated Dependencies and Dependencies
- Yocto 4.0 Kirkstone (was 3.1 Dunfell) (LTS)
  - [4.0 Kirkstone release notes](https://docs.yoctoproject.org/singleindex.html#release-notes-for-4-0-kirkstone)
  - [3.4 Honister release notes](https://docs.yoctoproject.org/singleindex.html#release-notes-for-3-4-honister)
  - [3.3 Hardknott release notes](https://docs.yoctoproject.org/singleindex.html#release-3-3-hardknott)

There are too many changes to fully list here; you can click the links and see all of them if you want to. Some major ones are noted here:
## General Stuff
- Kernel is following upstream better now
- Rust support in the toolchain (!)
- openssl 3.0.2 from the 1.1.l legacy release (finally)

## JS/ODD stuff
- Built in recipe for node 16 instead of backporting one
- Mesa 22
- I'm sure many more weird graphical things

## Python stuff
- Python 3.10 (!)
- Poetry is built in now (!)
- General build system improvements, including setuptools 59.5 and setuptools-rust-native; this should have better support for PEP-518 packages
- More built-in recipes for dependencies; we're not using these yet, but some stuff like jsonschema, etc is here now.
- python-can 4.1.0 (quite a lot in it - resumption of releases after a couple years of contributions but no publishing: https://github.com/hardbyte/python-can/releases )

# Changes besides bumping all the submodules

- Bumping from yocto 3.1 to yocto 4.0 is a lot of versions, and one of those version is 3.4 honister, in which [the syntax of variable overrides in recipes (blessedly) changed](https://docs.yoctoproject.org/singleindex.html#override-syntax-changes). There are therefore a lot of small touches to recipes to adapt the syntax. Most were executed with the script `<oe-core>/scripts/contrib/convert-overrides.py`.
- There are also the usual number of renamed, deprecated, and added variables that tended to get custom touches.
- Some machine configurations changed, and so did the style of toradex machine configurations. We had an overridden machine.conf, which was now just completely mismatched; I looked at the new one, removed everything that was exactly the same, and then dropped the remaining configuration that had actually changed into our distro conf. Those probably want to be machine-specific overrides somehow.
- for some reason we stopped being able to mkfs in the userfs backing file because it was too small; reading the mkfs manual it sure seemed like this would have always been a problem, and at the very least the userfs wouldn't have a journal, so I bumped it to a minimum size of 2MB which should be enough to always generate a journal (the journal is supposed to vary between 1MB and half the size of the filesystem, so I don't know if it'll be big enough, but whatever)

# Testing and reviews

This builds but I haven't yet even boot tested it yet. I guarantee that various things will not work. We need to try and run it and see what happens.


vegano1 (approved):

sfoster1:
- [ ] aiohttp needs to be at least version 3.8 to work under python 3.10

sfoster1:
- [ ] pythoncan needs to be >4.0.0 for 3.10 compat

sfoster1:
- [ ] weston doesn't seem to start because it's missing `/usr/lib/libweston-10/wayland-backend.so`

sfoster1:
CLosing in favor of #110

PR #44, opened 2023-01-06
fix(recipes-robot): we should not set proxy headers for the /protocols endpoint in nginx conf

Laura-Danielle:
## Overview

See RSS-166 for further issue on the bug found. Essentially, this change made the `/protocols` endpoint unusable. I am still debugging a separate issue that's popping up after this conf change so I will keep this PR open until I can find the root cause.


sfoster1 (approved):
lgtm! copy/paste mistake i think

vegano1:
Looks good,
tested /protocol GET/POST requests on OT3 devkit using the latest desktop app and postman.

mcous (approved):
🚢

vegano1 (approved):

PR #45, opened 2023-01-09
chore(build): Handle release builds when tags are pushed to monorepo

vegano1:
# Overview

The OT3 workflow needs to handle release builds, which are started when a tag is pushed to the monorepo in the form of "**ot3@**". In order to handle this we need to make some changes to how the workflow works and how the build is generated.

# Changes

- Updated the build-ot3-actions.yaml to check if a build is release based on the "ot3@" tag and set the build_type to release, develop accordingly.
- Set "OT_BUILD_TYPE" in conf/local.conf based on the build_type as determined by the check-release build step.
- Added release-notes.md to the list of generated files uploaded to s3, this is based on the latest /api/release-notes.md file of the monorepo. Also this will get posted to slack.
- remove "debug-tweaks" EXTRA_IMAGE_FEAURE by default and only set it based on the build_type==debug, so its removed for release builds.
- Added oe-version, oe-sha, oe-branch entries to the VERSION.json file
- Added scripts/update_releases_file.py to help update the releases.json file 

# Review Requests

- [x] Make sure we generate a release build when a tag matching "ot3@" is pushed to the monorepo
- [x] Make sure the generated VERSION.json file contains the correct build_type (release, develop)
- [x] Make sure the release-notes.md are posted hosted on S3 correctly, use the url in the [releases.json](https://ot3-development.builds.opentrons.com/ot3-oe/releases.json) file to confirm.
- [x] Make sure the releases.json file is updated when a new release is built, use the url [here](https://ot3-development.builds.opentrons.com/ot3-oe/releases.json) confirm.
- [x] Make sure the VERSION.json entries for oe-version, oe-sha, oe-branch are correct
- [x] Make sure we can boot after flashing the image to an OT3
- [x] Make sure we can boot after doing an in-line update from the latest edge build.

# Risk Assessment

Low, OT3 development build only


layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
```diff
@@ -60,16 +64,26 @@ MAX_SYSTEMFS_SIZE = "1536"
...
+        exit()
 
+    # Create the manifest dictionary
     opentrons_manifest = {
         'robot_type': d.getVar('ROBOT_TYPE'),
-        'build_type': os.getenv('OT_BUILD_TYPE', 'unknown/dev'),
-        'openembedded_version': d.getVar('version', 'unknown'),
-        'openembedded_sha': d.getVar('version', 'unknown'),
-        'openembedded_branch': d.getVar('version', 'unknown')
+        'build_type': d.getVar('OT_BUILD_TYPE', 'develop'),
```
sfoster1, line 83: we're using develop here as a default but the common values are `debug` and `release`. I like `develop` better than `debug` - can we use that as an intended value for `OT_BUILD_TYPE` also?

layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
```diff
@@ -107,6 +121,8 @@ do_make_rootfs_changes() {
 
     # add the VERSION.json file
     cat ${DEPLOY_DIR_IMAGE}/VERSION.json > ${IMAGE_ROOTFS}/etc/VERSION.json
+    # copy the release notes to the output dir
+    cat ${IMAGE_ROOTFS}/etc/release-notes.md > ${DEPLOY_DIR_IMAGE}/release-notes.md
```
sfoster1, line 125: maybe `${IMAGE_ROOTFS}${sysconfdir}`? that's what we're using in the robot server recipe

scripts/update_releases_file.py
```diff
@@ -0,0 +1,68 @@
...
+    releases_file = args.releases_file
+    version_file = args.version_file
+    base_url = args.base_url
+
+    # Get the version from the VERSION file
+    if os.path.exists(version_file):
+        with open(version_file, "r") as fh:
+            version = json.load(fh).get("update_server_version", "0.0.0-dev")
+    else:
+        print(f"ERROR: version file {version_file} not found!")
```
sfoster1, line 17: minor, but this should probably go to stderr i think?

.github/workflows/build-ot3-actions.yml
```diff
@@ -46,6 +46,18 @@ jobs:
...
           monorepo: ${{ inputs.monorepo-ref }}
           oe-core: ${{ inputs.oe-core-ref }}
+      - name: Check release build
+        id: check-release
+        if:
+          always()
+        run: |
+          echo "checking ${{steps.build-refs.outputs.monorepo}}"
+          build_type="develop"
+          if $(grep "ot3@" -q <<< ${{steps.build-refs.outputs.monorepo}}); then
```
sfoster1, line 56: i'm a little dubious about this because it relies on a second-order indirection - we're not parsing git tags, we're parsing the output of a github action that we wrote. What if instead of doing this we add it to the [action](https://github.com/Opentrons/oe-core/blob/main/.github/actions/build-refs/action/index.ts), where we can also add tests?

sfoster1 (changes requested):
I think we should push the build-type determination up into the typescript action, and I have some naming nits, but otherwise nice!

DerekMaggio (commented):
From what I can tell this looks okay. 
But that being said, there is some extremely dense logic going on in here. 

For instance, 
```yaml
      - name: Download sources
        run: |
          here=$(pwd)
          oe_mount="type=bind,src=$here/oe-core,dst=/volumes/oe-core,consistency=delegated"
          monorepo_mount="type=bind,src=$here/opentrons,dst=/volumes/opentrons,consistency=delegated"
          cache_mount="type=bind,src=${LOCAL_CACHE:-./cache},dst=/volumes/cache,consistency=delegated"
          echo "docker run --rm --mount $oe_mount --mount $monorepo_mount --mount $cache_mount ot3-image:latest opentrons-ot3-image --runall=fetch"
          docker run --rm --mount $oe_mount --mount $monorepo_mount --mount $cache_mount   ot3-image:latest opentrons-ot3-image --runall=fetch
      - name: Build image
        run: |
          here=$(pwd)
          oe_mount="type=bind,src=$here/oe-core,dst=/volumes/oe-core,consistency=delegated"
          monorepo_mount="type=bind,src=$here/opentrons,dst=/volumes/opentrons,consistency=delegated"
          cache_mount="type=bind,src=${LOCAL_CACHE:-./cache},dst=/volumes/cache,consistency=delegated"
          docker run --rm --mount $oe_mount --mount $monorepo_mount --mount $cache_mount ot3-image:latest opentrons-ot3-image
```

Could we maybe pull some of these steps out into callable actions inside of `.github/actions`?

See [Adding an action from the same repository](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#adding-an-action-from-the-same-repository)

vegano1:
> 

These two steps look dense but they are very straightforward, just that we have some very long arguments that we are separating into independent variables for readability. Ultimately this is just calling `docker run`. Ill add a comment so its clear whats happening here.

sfoster1:
yeah there's actually no logic there, it's really just a different method of separating things by line for readability

PR #51, opened 2023-02-03
recipes-robot/opentrons-mcu-firmware: add filepath from filename provided by subsystem_versions.py 

vegano1:
- fix an issue with wrong filenames
- Dont assume the names of the binary files, instead, have ot3-firmware repo's subsystem_versions.py script deal with that so oe-core can just deal with placing the files in the correct location.
- Change structure to `{"subsystem": {"branch": branch, "version": version, "shortsha": shortsha, "files_by_revision": {"rev1": filepath}}`


sfoster1 (approved):
Looks good!

PR #57, opened 2023-02-15
opentrons-ot3-image: Remove Toradex recipes that are causing mDNS discovery issues.

vegano1:
# Overview

The images are being built with Toradex utilities that are mainly for development/benchmarking purposes, one of these utilities (hostapd) provided by the **packagegroup-wifi-tdx-cli** recipe was causing mDNS discovery issues because the uap0 interface would disconect/reconnect every 2-3 minutes which caused avahi to stop broadcasting over mDNS.

# Changes

- Remove hostapd-example recipe.
- Disable uap0 interface

# Test Plan

- [x] Make sure we can boot the system
- [x] Make sure we can update this version
- [x] Make sure this does not break any CLI tools we are depending on
- [x] Make sure we can discover the OT3 after 5 minutes of boot without clearing desktop app cache
- [x] Make sure we can discover the OT3 after 5 minutes of boot after clearing desktop app cache

### DIFF
--- a/.github/workflows/build-ot3-actions.yml
+++ b/.github/workflows/build-ot3-actions.yml
@@ -108,18 +108,25 @@ jobs:
       - name: Pull S3 cache
         shell: bash
         run: |
-          aws_cp="aws --profile=${{ steps.aws.outputs.profile_name }} s3 cp --no-progress"
-          cachedir=${LOCAL_CACHE:-./cache}
-          for cachetype in downloads sstate git ; do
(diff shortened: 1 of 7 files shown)


sfoster1 (commented):
[leaving comment instead of request-changes to avoid blocking]
We for sure want to keep `wireless-regdb-static`, or at least supply a replacement, otherwise wifi will stop working outside the US. maybe add that back in and keep the packagegroup removed.

vegano1:
> 

good point

sfoster1 (approved):
Nice!

PR #58, opened 2023-02-16
recipes-images/opentrons-ot3-image: add dfu-util to support rear panel board updates.

vegano1:
# Test Plan

- [x] Make sure **dfu-util** cli command works on OT3


sfoster1 (approved):
looks good if it works

PR #59, opened 2023-02-21
recipes-images/opentrons-ot3-image: Increase rootfs size to 2GB + check max rootfs size when building.

vegano1:
# Overview

We have a lot of stuff on the OT3, and the 1.5GB of space we have for system services has already been used up. However, the builds were not taking this into account and were generating images that were bigger than the partition size they were to be installed in. The system-server also does not check if the update size fits in the partition it's writing to, which caused critical files to be overwritten, and when the system restarts that would brick it. 

So first let's increase the size of the partition to 2GB, this won't get us very far but it will at least get us moving.
Secondly, we need not produce images that are larger than the partition size so let's take that into account and bail.
Ideally, we can reduce the size of our packages, but let's stop the bleeding first.

# Changes
- Increase systemfs size to 2GB
- Don't build an image if the systemfs (rootfs) is too large for the systemfs partition.

# Test Plan

- [x] Make sure we dont build an image if the size is too large
- [x] Make sure generated image is now 2GB
- [x] Flash on OT3 and make sure we can boot

### DIFF
--- a/layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
+++ b/layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
@@ -65,7 +65,7 @@ IMAGE_NAME = "${MACHINE_NAME}_${IMAGE_BASENAME}"
 USERFS_DIR = "${WORKDIR}/userfs"
 USERFS_OUTPUT = "${DEPLOY_DIR_IMAGE}/userfs.ext4"
 # max rootfs partition size in mb
-MAX_SYSTEMFS_SIZE = "1536"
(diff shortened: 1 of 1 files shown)


fsinapi (approved):
LGTM if the build goes through, and assuming you were able to get the error to trigger by keeping the `MAX_SYSTEMFS_SIZE` to the old value. Love to see an awk script.

vegano1:
> LGTM if the build goes through, and assuming you were able to get the error to trigger by keeping the `MAX_SYSTEMFS_SIZE` to the old value. Love to see an awk script.

correct

PR #61, opened 2023-03-03
build: use an explicit tmp volume to fix out of disk problems

sfoster1:
We've recently been having some issues with builds stopping because diskmon says we're out of disk space. In response, we limited the size of the cache we'd download (#52 ). 

However, it kept happening occasionally: https://github.com/Opentrons/oe-core/actions/runs/4325480403/jobs/7551693531

It fails with this message:
```
WARNING: The free space of /tmp (overlay) is running low (0.049GB left)
ERROR: No new tasks can be executed since the disk space monitor action is "STOPTASKS"!
```

Though we do have a build directory called `tmp`, that's not what this is: this is about the actual `/tmp` mounted inside the build container. The thing is, `/tmp` should be a special `tmpfs` mount, a ramdisk that can page if necessary and have effectively unlimited size.

However, when we set up the image, we mount an nvme volume into `/`. That appears to stomp on the normal `/tmp` mount, and that gets mapped into the container, and _that_ runs out of space.

We can fix this pretty easily! Docker on linux lets you specify a `tmpfs` mount type in `run`, and that is a real `tmpfs` that really is infinite size, and then when we do that we can reenable caching, and then builds are way faster again.

The build for this PR is https://github.com/Opentrons/oe-core/actions/runs/4325932098

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -33,5 +33,7 @@ patch -f ./layers/meta-toradex-nxp/recipes-kernel/linux/linux-toradex_5.4-2.3.x.
 export BITBAKEDIR=${THISDIR}/tools/bitbake
 . layers/openembedded-core/oe-init-build-env ${THISDIR}/build
 
+df -h /tmp
+
 BB_NUMBER_THREADS=$(nproc) bitbake ${TARGET} "$@"
 exit $?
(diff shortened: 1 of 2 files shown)


start.sh
```diff
@@ -33,5 +33,7 @@ patch -f ./layers/meta-toradex-nxp/recipes-kernel/linux/linux-toradex_5.4-2.3.x.
 export BITBAKEDIR=${THISDIR}/tools/bitbake
 . layers/openembedded-core/oe-init-build-env ${THISDIR}/build
 
+df -h /tmp
```
vegano1, line 36: debug print?

vegano1:
This lgtm, thank you

vegano1 (approved):

PR #64, opened 2023-03-13
opentrons-mcu-firmware: allow .bin for rear panel

sfoster1:
The rear panel system recently changed to have its applications use .bin instead of .hex for the files; this makes OE complain because we told it to expect us to install only .hex. Add the rear-panel binaries to FILES.


vegano1 (approved):
lgtm

PR #65, opened 2023-03-13
recipes-robot/nginx: add system server carveout to nginx.conf

fsinapi:
This PR adds configuration to NGINX to allow external access to the system server.

Also adds an explicit config to route `/system/time` to the robot server, as that endpoint is not ported to the system server yet.

Tested by uploading the edited `nginx.conf` to a robot and running `nginx -s reload`, and also uploading the system server changes from https://github.com/Opentrons/opentrons/pull/12278. 
* I was able to go to `{robot_url}:31950/system/docs` and interact with the endpoints on the system server without issue. 
* I was also about to go to `{robot_url}:31950/docs` and interact with `GET /system/time` without issue.

### DIFF
--- a/layers/meta-opentrons/recipes-robot/nginx/files/nginx.conf
+++ b/layers/meta-opentrons/recipes-robot/nginx/files/nginx.conf
@@ -50,5 +50,21 @@ http {
             proxy_request_buffering  off;
             client_max_body_size     0;
         }
+
+        location /system {
+            proxy_http_version       1.1;
+            proxy_read_timeout       1h;
+            proxy_pass               http://127.0.0.1:32950;
+        }
+
+        # The /system/time endpoints live in the robot server, and thus
+        # the configuration here has to match the / config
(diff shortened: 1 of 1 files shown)


layers/meta-opentrons/recipes-robot/nginx/files/nginx.conf
```diff
@@ -50,5 +50,21 @@ http {
...
+
+        location /system {
+            proxy_http_version       1.1;
+            proxy_read_timeout       1h;
+            proxy_pass               http://127.0.0.1:32950;
+        }
+
+        # The /system/time endpoints live in the robot server, and thus
+        # the configuration here has to match the / config
+        location /system/time {
```
vegano1, line 62: same as the buildroot pr,
We should be able to nest location `/system/time` in the root config
everything else looks good though

sfoster1 (approved):
Looks good to me!

PR #66, opened 2023-03-16
opentrons-robot-app: allow remote devtools

sfoster1:
Chrome has a [remote devtools protocol](https://chromedevtools.github.io/devtools-protocol/) that you can use to separate where you run devtools from where you're running content - pretty similar to a remote debugging interface like `gdb-server` or the mac tandem kernel debugger or whatever. That's really convenient because then we don't have to try and use the devtools on the little touchscreen.

Since electron is chrome, it's built into electron, and you can tell electron to listen on a port. The problem is, electron will only listen on that port _to localhost_, and the entire point of this exercise is to make this accessible off of localhost.

As always, systemd has a solution for this: [systemd-socket-proxyd](https://manpages.ubuntu.com/manpages/bionic/man8/systemd-socket-proxyd.8.html) is a socket proxy (works on tcp and udp, used here for tcp) that can be activated by a [systemd socket unit](https://manpages.debian.org/testing/systemd/systemd.socket.5.en.html) that listens on a port. They can't use the same port without more extensive configuration (read: a bunch of bindings to come up and down as interfaces do, since binding to 0.0.0.0:nnnn conflicts with binding to 127.0.0.1:nnnn), so we can expose 9223 for an internal connection to 9222, and now you can point a chrome at it and get devtools! hooray!

### DIFF
--- a/layers/meta-opentrons/recipes-robot/opentrons-robot-app/files/opentrons-robot-app-devtools.service
+++ b/layers/meta-opentrons/recipes-robot/opentrons-robot-app/files/opentrons-robot-app-devtools.service
@@ -1,0 +1,10 @@
+[Unit]
+Requires=opentrons-robot-app.service
+After=opentrons-robot-app.service
+Requires=opentrons-robot-app-devtools.socket
+After=opentrons-robot-app-devtools.socket
+JoinsNamespaceOf=opentrons-robot-app.service
+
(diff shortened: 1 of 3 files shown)


vegano1 (approved):

Laura-Danielle (approved):
cool

PR #67, opened 2023-03-16
meta-opentrons: add splash screens during loading

sfoster1:
When the system is booting, we want to display splash screens or sequences that are under our control rather than the default ones that come with openembedded. 

The ideal situation is to have a different screen for each part of the boot process; that way, if a part of the boot process fails, you can tell which part from a simple picture of the screen. We can't do this for the bootloader because u-boot doesn't have drivers for the DSI bridge we use, but we can for the kernel and early userspace.

These get implemented in two different ways:

### Kernel
The kernel has some splash screen functionality that works by compiling in a bitmap in a special format. Instructions for how to do so are [available from toradex](https://developer.toradex.com/linux-bsp/os-development/boot/splash-screen-on-the-linux-kernel/). To get this to display appropriately though, we need to make sure the kernel isn't displaying anything else. Specifically, that means turning off the video console that the kernel is currently configured with, which is a boot arguments change. So https://github.com/Opentrons/oe-core/commit/a00652b7a67530e339810fb1363061bd79b83d24 and https://github.com/Opentrons/oe-core/commit/a57c4b571219c8733fbdbeb758bad4b807834489 remove the default of these args from the u-boot environment, and then change the way we pass those args because it wasn't working anyway.

The actual kernel logo is added in https://github.com/Opentrons/oe-core/commit/00cba809547e73c9257cf297daefe83f4c90cd41 , and there's really no more work needed than bootargs + logo availability.

Finally, since we were editing boot args anyway, let's toss the modesetting arguments in to the actual boot script; they'd previously been set in the wrapup script of the easyinstaller, which is pretty hard to find: https://github.com/Opentrons/oe-core/commit/8535c0816ce8ad3c634f57e00185570cb820868c

### Userspace

Userspace is a little more involved to get working, because we want an animation and because we want the loading screen to coincide with the app starting and stopping.

The animation issue is solved by requiring the "animation" to actually just be an mp4. Gstreamer can play those. We use gstreamer-launch to view one-off videos all the time. However, it's a real pain to get gstreamer to seamlessly loop videos from a shell script - we'd have to write a program to do that - so we use [gstd](https://developer.ridgerun.com/wiki/index.php/GStreamer_Daemon_-_GStreamer_Daemon_Basics), which runs in the background and does all that for us. That way we can take a short mp4 and loop it to make it look like a continuous animation.  It would be nicer if it was a gif, but for some reason I cannot figure out, gstreamer doesn't have gif decode support. Mp4 is pretty compressed anyway.

Once we can display animations, we need to display them at the right times. The `opentrons-loading` service can do that with appropriate systemd hooks to start as soon as possible after weston and end when the app starts. It can also pick back up if the app crashes.

### What's left?
There's still a couple seconds of no-display in between the end of the kernel logo and the beginning of the animation. This is because that's before weston starts. There's probably a dance we could do with one kind of loading screen running from early boot to weston, then a second running after weston... I think that's on the wrong side of the 80:20 curve.

Closes RCORE-348


Laura-Danielle (approved):
Should definitely have Brayan review this, but from what I can tell lgtm.

PR #68, opened 2023-03-17
.github: fix failure looking up tags

sfoster1:
When we look up the latest tag, we were trying to find the literal ':latest:' in the list of tags we got back from github, which won't ever work. Instead add a tested helper function for that.


sfoster1:
Build showing correct behavior where we get all tags for all the repos: https://github.com/Opentrons/oe-core/actions/runs/4451705408/jobs/7818657989

shlokamin (approved):
🏷️

PR #70, opened 2023-03-22
opentrons-robot-server: use internal release notes

sfoster1:
These will be set for internal releases. We'll have to change this before full production release.


vegano1:
Looks good,
can we add a todo comment so we know that has to change

sfoster1:
Can be viewed at ot3-development.builds.opentrons.com/ot3-oe/4501787501/release-notes.md

vegano1 (approved):

PR #73, opened 2023-03-27
recipes-robot/opentrons-user-environment: add OT_SYSTEM_VERSION environment flag

vegano1:
We need to set the OT_SYSTEM_VERSION environment variable so the `opentrons` module can be imported from jupyter notebook 

* Test on the Flex
* Make sure the `OT_SYSTEM_VERSION` env var is set properly
* Make sure we can import `opentrons` package from jupyter notebook

### DIFF
--- a/layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
+++ b/layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
@@ -120,6 +120,12 @@ python do_create_opentrons_manifest() {
     opentrons_json_output = "%s/VERSION.json" % d.getVar('DEPLOY_DIR_IMAGE')
     with open(opentrons_json_output, 'w') as fh:
         json.dump(opentrons_manifest, fh, indent=4)
+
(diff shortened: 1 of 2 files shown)


layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
```diff
@@ -120,6 +120,12 @@ python do_create_opentrons_manifest() {
     opentrons_json_output = "%s/VERSION.json" % d.getVar('DEPLOY_DIR_IMAGE')
     with open(opentrons_json_output, 'w') as fh:
         json.dump(opentrons_manifest, fh, indent=4)
+
+    # write the system version file
+    ot_system_version = "%s/OT_SYSTEM_VERSION" % d.getVar('DEPLOY_DIR_IMAGE')
```
sfoster1, line 125: The thing is, this fully duplicates an element of `/etc/VERSION.json`. I don't really love adding the extra file.

PR #75, opened 2023-04-21
opentrons-robot-server: longer start timeout

sfoster1:
This is really annoying! But we need to have a long enough timeout for now to allow for all the firmware updates.


fsinapi (approved):
🔟

ryanthecoder (approved):
lgtm

PR #77, opened 2023-05-19
build: move yarn cache location to mounted volume when building from CI

vegano1:
yarn and node-modules were taking up a ton of space inside the docker container causing us to run out of space, so let's move the .cache location that yarn uses to cache modules to /volumes/cache/yarn which gets bind mounted outside the containers in CI.

image


sfoster1 (approved):
Thank you!

PR #78, opened 2023-05-30
recipes-kernel: add M24128 EEPROM overlay to device tree

vegano1:
We need to add the m24128 eeprom we are using in the Opentrons carrier board to the device tree so we can access the memory from userspace.

Closes: [RET-1345](https://opentrons.atlassian.net/browse/RET-1345)

Todo:
- [x] Make sure the M24128 eeprom is initialized on start-up and shows up on dmesg
- [x] Make sure we can read/write to /sys/bus/i2c/devices/3-0050

[RET-1345]: https://opentrons.atlassian.net/browse/RET-1345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfoster1 (approved):
looks good if it works!

PR #79, opened 2023-06-02
build(fix): move electron cache dir to bind mounted for in CI

vegano1:
electron uses a separate cache location to that of yarn, which was causing yet another running out of space issue.


sfoster1 (approved):
Nice, thank you! Glad there was an env var for this.

PR #80, opened 2023-06-06
opentrons-module-firmware: added module firmware binaries to rootfs.

vegano1:
A recipe that downloads and installs the module firmware binaries to `/usr/lib/firmware` directory to enable firmware updates for modules.

# Test Plan

- [x] Install the image on flex and make sure the module binaries are in `/usr/lib/firmware`
- [x] Make sure out-of-date modules are updated to the correct firmware by the robot server when connected to a flex.


sfoster1 (approved):
nice!

PR #81, opened 2023-06-14
layers/meta-opentrons: save logs on disk

sfoster1:
We previously had storage on auto but didn't have a /var/logs/journal so logs were going only to RAM with a limit of 64 megs. Now we'll make the directory and store 4G of logs.


Laura-Danielle (approved):
nice

ahiuchingau (approved):
🙏

sfoster1:
```
cat mount/lib/systemd/journald.conf.d/01-storage.conf
[Journal]
Storage=persistent
ForwardToSyslog=yes
```
it (finally) works with the explicit recipe override

vegano1:
I was concerned about the 4GB max journal size this will take up, but by default, it's either 15% of the total filesystem space OR 4GB. Since this is in the userfs partition it will be fine.

PR #84, opened 2023-06-27
fix(build): create symlink for electron cache dir manually

vegano1:
The `electron_cache_dir` env variable is being ignored when set from opentrons-robot-app recipe, so lets just manually create a symlink from `~/.cache/electron` -> `/volumes/cache/electron` which bind mounts to a larger partition in the ec2 instance.


vegano1:
waiting on build to finish confirm change

sfoster1 (approved):
Yup yup!

PR #86, opened 2023-07-07
opentrons-systemd-units: fix machine-id being re-generated after reboot

vegano1:
## Overview 

The `/etc/machine-id` is used by systemd for a few different things, including serving as the unique identifier for journald logs. This usually persists through power cycles and updates, but our rootfs is read-only so we need to do some workaround to get it to persist. 

This pr adds that workaround, by adding a systemd service that takes over for the [systemd-machine-id-commit.service](https://www.freedesktop.org/software/systemd/man/systemd-machine-id-commit.service.html#). Our service bind-mounts the `/etc/machine-id` file in a temporary read-write location at which point we write the machine-id which writes to the underlying machine-id file, making it persist over reboots and power cycles. See the ticket below for more details

Closes: [RCORE-706](https://opentrons.atlassian.net/browse/RCORE-706)

## Test Plan

- [ ] Flash the image on the robot and make sure the `/etc/machine-id` persists through power cycles
- [ ] Update the robot OTA and make sure the `/etc/machine-id` persists

### DIFF
--- a/layers/meta-opentrons/recipes-robot/opentrons-systemd-units/files/opentrons-commit-machine-id.service
+++ b/layers/meta-opentrons/recipes-robot/opentrons-systemd-units/files/opentrons-commit-machine-id.service
@@ -1,0 +1,6 @@
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/ot-commit-machine-id
+
+[Install]
+WantedBy=basic.target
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
Looks good! works on the ot-2, should work here!

SyntaxColoring (approved):
Yay, thanks.

Hypothetically, is there a decent way to deduplicate stuff like this between https://github.com/Opentrons/buildroot and https://github.com/Opentrons/oe-core?

vegano1:
> Yay, thanks.
> 
> Hypothetically, is there a decent way to deduplicate stuff like this between https://github.com/Opentrons/buildroot and https://github.com/Opentrons/oe-core?

Yes, ideally we migrate the OT2 to Yocto then we can share recipes between machines.

PR #87, opened 2023-07-11
recipes-core/base-files: mount the /mnt dir to /userfs/var/mnt so the machine-id can persist through updates.

vegano1:
Works with [this](https://github.com/Opentrons/opentrons/pull/13077) monorepo change to fix the machine-id not persisting through updates.


sfoster1 (approved):
oh yeah that's a good idea good call

PR #88, opened 2023-07-13
build(fix): Manually set yarn download dir to cached location.

vegano1:
here we go again...

We should be setting the yarn download dir to /volumes/cache/yarn within the opentrons-robot-app.bb recipe, but for some reason that's being ignored, so let's just manually set the download dir with a symlink as we do for electron.


sfoster1 (approved):
Looks good to me!

PR #89, opened 2023-07-13
meta-opentrons: robot-app: remove cypress

sfoster1:
Cypress is gigantic and downloads to the wrong place. Preemptively remove it from the node dependencies since it's not necessary for building the actual app.


vegano1 (approved):
cya

PR #90, opened 2023-07-19
build: set OPENTRONS_PROJECT from workflow

sfoster1:
We can do this in the localconf. Soon, we're going to make a version of this workflow that uses the robot stack version numbers (e.g. 7.0.0) and we'll need to set this variable to `robot-stack` accordingly.

## Testing
- [x] the build should be the same as before
  - [x] specifically the version file in the rootfs should have all the right 0.13.0 or whatever tags
  - [x] so should the python modules
  - [x] so should the ODD


vegano1 (approved):
This makes sense to me

PR #91, opened 2023-07-24
opentrons-ot3-image: Add rootfs hash signing to the build

vegano1:
# Overview

We need to make sure our builds are official Opentrons builds before installing the rootfs, the logic that verifies the rootfs hash and hash signature are already on the monorepo. This pr adds the plumbing for producing builds that contain the signed rootfs hash which is used by the monorepo to verify the build.

Closes: [RCORE-309](https://opentrons.atlassian.net/browse/RCORE-309)

# Test plan

- [x] Make sure we can install this build from a build without the Opentrons cert key.
- [x] Make sure that once installed we have a `/etc/opentrons-robot-signing-key.crt` file on the rootfs.
- [x] Make sure that OTA updates via the update-server now reject non-signed builds.
- [x] Make sure that OTA updates via the update-server are installed when signed.

Note: The update-server checks the existence of the cert key `/etc/opentrons-robot-signing-key.crt` on the first boot, if the file is missing the update-server sets the `signature_required` key to false in the update-server config file at `/var/lib/otupdate/config.json`. The `signature_required` key determines if the update-server verifies the build using the cert key, so for all builds until now we would need to remove the `/var/lib/otupdate/config.json` file so it gets recreated with the `signature_required` key set to true.

[RCORE-309]: https://opentrons.atlassian.net/browse/RCORE-309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfoster1 (changes requested):
> Note: The update-server checks the existence of the cert key `/etc/opentrons-robot-signing-key.crt` on the first boot, if the file is missing the update-server sets the `signature_required` key to false in the update-server config file at `/var/lib/otupdate/config.json`. The `signature_required` key determines if the update-server verifies the build using the cert key, so for all builds until now we would need to remove the `/var/lib/otupdate/config.json` file so it gets recreated with the `signature_required` key set to true.

One missing thing: the above behavior is intentional to have "dev builds" that do not check that the update they're sent is signed - so you can still make dev builds and install them on a robot. We need to build that behavior into the automated builders by removing the certificate if the build is not a release build. On these dev builds, we won't have a certificate, the update server will notice, and then signatures won't be required.

At least, that's how we do it on the OT-2. If you want, we could also override the setting on dev builds somehow - whatever you think is the cleanest way of achieving this.

sfoster1 (approved):
Loooks good too me!

PR #92, opened 2023-07-27
fix(build): add OPENTRONS_PROJECT var to local.conf so we can do local builds.

vegano1:
The OPENTRONS_PROJECT var is set by CI during a build, but nothing sets it for local builds so the Opentrons python packages that rely on it fail to compile.


sfoster1 (approved):
Oops, sorry! This gets overridden appropriately in CI still right?

yep, it does,
CI appends this to the end of the file, which overrides the variable.

PR #95, opened 2023-08-04
Build: handle release builds

dapiroy:
Updated workflow for release builds to handle both release and internal release


.github/workflows/build-ot3-actions.yml
```diff
@@ -30,13 +30,38 @@ on:
         default: 'stage-prod'
 
 jobs:
+  decide-refs:
```
vegano1, line 33: This looks like it can be done below when we also call "decide-refs" since all we are doing is adding the variant variable.

.github/workflows/build-ot3-actions.yml
```diff
@@ -277,13 +302,21 @@ jobs:
...
-      - name: Post results
-        if: ${{ matrix.build_env == 'stage-prod' }}
+      - name: Post results as internal-release
+        if: ${{ matrix.build_env == 'stage-prod' && steps.build-refs.outputs.variant=='internal-release' }}
         uses: slackapi/slack-github-action@v1.14.0
         with:
           payload: "{\"s3-url\":\"${{ steps.upload-results.outputs.console_url }}/\",\"type\":\"branch\", \"reflike\":\"${{ steps.build-refs.outputs.oe-core }}\",\"monorepo-reflike\":\"${{ steps.build-refs.outputs.monorepo }}\",\"firmware-reflike\":\"${{ steps.build-refs.outputs.ot3-firmware }}\",\"full-image\":\"${{ steps.upload-results.outputs.fullimage_url }}\", \"system-update\":\"${{ steps.upload-results.outputs.system_url }}\", \"version-file\":\"${{ steps.upload-results.outputs.version_file_url }}\", \"release-notes\":\"${{ steps.upload-results.outputs.release_notes_file_url }}\"}"
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+
```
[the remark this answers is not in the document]
sfoster1: They're not - that bucket is provided by the runner and depends on the infra stage

.github/workflows/build-ot3-actions.yml
sfoster1: it needs to be done here because the subsequent job will run in different infrastructure depending on its output. however, we can remove the one below.

sfoster1: (and did so)

vegano1 (approved):
Ok, this looks good now.

PR #97, opened 2023-08-17
chore: add mixpanel for ODD

sfoster1:
Adds the mixpanel API key for upstreaming events when on a release build (intermediate build where it's always present for testing).


vegano1 (approved):
Looks good!

sfoster1:
The version of this branch that always adds the mixpanel id instead of just during release works great!

PR #100, opened 2023-09-06
opentrons-ot3-image: set the rootfs size to a fixed value when creating the system update image.

vegano1:
The rootfs being produced when creating the systemfs.ext4 filesystem was not set to a fixed size, which meant that the ext4 size varied based on what recipes were built. This has the effect of producing an image for flashing with the correct rootfs partition size of 2GB, but an ot3-system update file of varying size. Which when installed would set the new rootfs size to the varying size since the filesystem was not padded. So let's use the `IMAGE_ROOTFS_SIZE` variable to set the maximum rootfs size along with `IMAGE_OVERHEAD_FACTOR` and `IMAGE_ROOTFS_EXTRA_SPACE` so there is no additional extra space in the image other than what we've manually specified.

Also, since we delegate the creation of the rootfs (systemfs) to Yocto we don't need to have an explicit mechanism to check if the size of the rootfs is going to go over our set limit, so let's just remove that.

### DIFF
--- a/layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
+++ b/layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
@@ -208,7 +208,7 @@ python do_create_tezi_manifest(){
     tezi_manifest = {}
 
     # define the ot3  partitions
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
This makes sense to me, a lot less complex too

PR #101, opened 2023-09-11
base-files: added /media to custom fstab so we can auto-mount thumbdrives.

vegano1:
The `/etc/udev/scripts/mount.sh` script mounts usb thumb drives to "/media" but since the rootfs is read-only we cant, so lets bind-mount /media to the `/userfs/media` read-write partition.


sfoster1 (approved):
noice

PR #102, opened 2023-09-12
recipes-core/dropbear: add custom dropbear config to enable ssh with rsa keys

vegano1:
# Overview

For development, we have kept SSH RSA keys disabled, which means you can SSH to the Flex without using a password or private key. Since we are getting ready to ship the Flex we need to at least make it so you can't ssh as root without providing a valid private ssh key. This PR changes the config we provide to dropbear which disabled root logins with passwords, so the only way to be able to SSH into the Flex is with a valid private key for which the Flex has a matching public key.

# Test Plan

- [x] Using the new `/server/ssh_keys/from_local` endpoint add a valid SSH public key to the Flex and make sure you can SSH using the matching private key
- [x] Make sure you CAN'T SSH to a Flex if you don't provide a valid SSH private key.
- [x] Make sure development builds still don't require an SSH private key
- [x] Make sure the public key matches the private key (provided privately)

# Changes

- Added a custom Dropbear config file that sets `DROPBEAR_EXTRA_ARGS="-s"` to enable SSH rsa logins ONLY
- Custom Dropbear config is enabled only for release builds, so development still doesn't require ssh keys
- Added default Opentrons SSH public key


sfoster1 (approved):
Looks good to me if it works!

PR #104, opened 2023-09-16
opentrons-ot3-image: Add python3 pip module so users can install python packages.

vegano1:
# Overview

We want users to be able to install custom python packages using pip, so let's add that module and configure it so it uses the `/userfs/var/` directory when installing packages.

# Changelog

- Add python3-pip to opentrons-ot3-image recipe
- Add custom python3-pip action in do_install to install a custom pip.conf and .pth file to set install dir
- Add `/var/user-packages/usr/bin` to PATH env variable so we can access installed executables.

# Test

- [x] Make sure pip was installed by running `pip3`
- [x] Install a python package not included with the system by running `pip3 install gspread`
- [x] Make sure the package is installed to `/var/user-packages/usr/lib/python3.8/site-packages/`
- [x] Make sure we can import this package from the Python interpreter
- [x] Make sure we can import this package from a Jupyter Notebook journal


sfoster1 (approved):
ice, looks good to me!

PR #107, opened 2023-09-20
opentrons-ssh-keys: always install file

sfoster1:
We weren't installing the key in the release execution branch but were still providing it in the OE recipe. At some point we should make the build mode and project actual OE build variables that you can establish recipe variants on but for now just put an empty file in there in the other branch.


vegano1 (approved):
oops

PR #108, opened 2023-09-25
chore(build): Improve build caching so subsequent builds dont take so long.

vegano1:
- Don't clear cache (downloads, sstate, git) unless the build explicitly fails and is not just canceled.
- Create symlink for pip cache so it persists through builds


sfoster1 (approved):
Looks good to me!

PR #109, opened 2023-09-29
fix: make builds go fast again

sfoster1:
A recent pip/setuptools update made pep517 isolated builds the default. That's really good and I would like to use it but also it makes the builds incredibly slow: https://github.com/pypa/pip/issues/10060 We need to keep pep517 because it's what allows pyproject tomls to specify build deps, but by preinstalling those build deps and turning off build isolation we can make it fast again. This isn't a good general purpose solution and it will surely break at some point when we add some new package but we can cross that bridge when we come to it


vegano1 (commented):
Works for now 🙏

vegano1 (approved):

sfoster1:
Uploaded to a robot, looks pretty good!

PR #112, opened 2023-11-16
recipes-robot/opentrons-jupyter-notebook: add env variables to fix jupyter notebook.

vegano1:
# Overview

The Jupyter Notebook service did not have all the correct environment variables, so we could not get a protocol context instance. This PR adds the correct environment variables to the Jupyter Notebook systems file, since we are using the OT_SYSTEM_VERSION variable in multiple recipes we also added a helper bbclass to get this value.

Closes: [RQA-1258](https://opentrons.atlassian.net/browse/RQA-1258)

# Test plan

- [x] Install the image on the robot and make sure the jupyter notebook service starts up
- [x] Make sure we can import `opentrons.execute` from Jupyter
- [x] Make sure we can get a protocol context instance from Jupyter 
`protocol = opentrons.execute.get_protocol_api('2.15')`
- [x] Make sure we can call `procotol.home()` from jupyter and that it actually homes the gantry.

[RQA-1258]: https://opentrons.atlassian.net/browse/RQA-1258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/layers/meta-opentrons/recipes-robot/opentrons-jupyter-notebook/files/jupyter-notebook.service
+++ b/layers/meta-opentrons/recipes-robot/opentrons-jupyter-notebook/files/jupyter-notebook.service
@@ -11,6 +11,8 @@ SupplementaryGroups=dialout
 StateDirectory=jupyter jupyter/notebooks jupyter/data jupyter/config ipython
 RuntimeDirectory=jupyter
 Environment=RUNNING_ON_VERDIN=true
+Environment=OT_SYSTEM_VERSION=
+Environment=OT_API_FF_enableOT3HardwareController=true
 Environment=PYTHONPATH=/opt/opentrons-robot-server:/usr/lib/python3.8/site-packages
 Environment=JUPYTER_RUNTIME_DIR=%t/jupyter
 Environment=JUPYTER_CONFIG_DIR=%S/jupyter/config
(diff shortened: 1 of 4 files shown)


layers/meta-opentrons/recipes-robot/opentrons-jupyter-notebook/files/jupyter-notebook.service
```diff
@@ -11,6 +11,8 @@ SupplementaryGroups=dialout
 StateDirectory=jupyter jupyter/notebooks jupyter/data jupyter/config ipython
 RuntimeDirectory=jupyter
 Environment=RUNNING_ON_VERDIN=true
+Environment=OT_SYSTEM_VERSION=
```
sfoster1, line 14: nit: when doing this sed replacement style i kind of like having a placeholder value to replace, since this looks like it's trying to explicitly clear the system version. Can we make it like `OT_SYSTEM_VERSION=@@OT_SYSTEM_VERSION@@` or something similar and replace that?

sfoster1 (changes requested):
Looks good functionality wise but let's make it a little more clear when reading the service file that we're going to sed in a value

sfoster1 (approved):
Looks good to me, thanks!

PR #113, opened 2023-11-20
recipes-extended/procps: Enable magic SysRq keys to debug potential kernel panics.

vegano1:
# Overview

One of our robots was found unresponsive after the weekend, and could not connect over serial. We think it might be a resource allocation issue but can't verify yet, so let's enable magic SysRq keys so we can debug this when/if it happens again.

# Testing

- [x] Make sure magic keys are enabled by issuing reboot command


sfoster1 (approved):
Thank you! Looks good to me if it works

PR #114, opened 2023-12-08
meta-opentrons: add linklocal config for eth0

sfoster1:
We need to support falling back to link-local connections for initial installs. The way you do this with network manager is to preconfigure connections for the eth0 iface, one higher-priority one that uses dhcp and can fail and one lower-priority one that uses linklocal. These need to go into /var/lib/NetworkManager/system-connections, but of course that's not on the rootfs, so they need to get copied at boot by a custom unit.

You'll know this works if you can install this build, plug it into a computer directly, and after 30 seconds you get a link local address.


sfoster1:
Tested and works!

fsinapi (approved):
cool and good!

ahiuchingau (approved):
noice

PR #115, opened 2023-12-08
meta-opentrons: allow s-d setup-js to fail

sfoster1:
We recently removed setup-js from the shared-data makefile because it's not needed anymore, but the recipe for the ODD needed to know that. Temporarily just allow it to fail instead of removing it since sometimes it will be needed until we're forever done building old branches.


fsinapi (approved):
Nice. What's the over/under on how long this shim is in here?

PR #116, opened 2024-01-02
fix(build): Create symlink to .cache so we dont run out of space

vegano1:
Some Node packages are not using the correct cached directory for downloads, leading the docker instance to balloon in size causing the builds to fail. Instead of trying to wack-mole each new package whenever a build fails, let's symlink all of the `/volumes/cache` used by the docker instance to `~/.cache` which is externally mounted to S3 and has enough space to hold our cache.


sfoster1 (approved):
Looks good to me if it all works!

PR #117, opened 2024-01-02
chore: allow internal-release tags

sfoster1:
Add support for internal-release tags on ot3-firmware and oe-core. Internal release tags on those repos are formatted as internal@version. Builds that start from internal-release tags on the monorepo or here will use internal-release tags from everything instead of release tags.


vegano1:
This makes sense, lgtm.

vegano1 (approved):

PR #120, opened 2024-01-18
chore:  bump the pyzmq version and update the rpath patch

ryanthecoder:
we need an updated pyzmq for the updated jupyter notebook


sfoster1 (approved):
Looks good to me if it works, but let's get a successful build installed to confirm

vegano1 (approved):
I like

PR #123, opened 2024-01-22
Add plymouth

sfoster1:
toradex 6.0 upgrades to kernel 5.15 and does not support kernel bootsplash anymore. Instead, you're supposed to use plymouth ([freedesktop plymouth docs](https://www.freedesktop.org/wiki/Software/Plymouth/), [tdx integration docs](https://developer.toradex.com/linux-bsp/os-development/boot/how-to-add-custom-splash-screen/) ). So, 
- add plymouth
- add a custom theme that displays the same image as before (though now it can be a png instead of having to be ppm'd)
- remove the old kernel customization

This does work on both first-flash and system upgrade paths but there's some interesting things about it, like
- Plymouth is really intended to run in an initrd, ideally one managed by dracut. We don't use either of those things, which means that this implementation has a later boot splash than when it was managed by the kernel
- Because of this, plymouth is not capable of loading arbitrary paths and thus this is not really more customizable than the kernel splash
- plymouth is essentially undocumented aside from random blog posts linked at the bottom of the freedesktop docs and random archwiki posts
- Rather than loading many kinds of media (which would require a lot of code), it has a custom scripting language that lets you do sprites to do loading spinners or whatever, which means we can't actually integrate our two bootsplashes still
- If you get anything wrong in its configuration plymouth segfaults
- Command line flags and configuration file options are largely undocumented and can only be figured out from reading the code.
- There are still too many modeswitches and therefore black screens including when weston takes over, and not having any of these is the literal entire point behind plymouth's existence, but I'm pretty sure that's weston's fault and will be a bit of a rabbit hole
- All of the above can be true and we still need to do this because kernel bootsplash just straight up doesn't exist anymore sorry   ¯\_(ツ)_/¯


vegano1 (approved):
Thank you!
Animating is going to be fun :)

PR #124, opened 2024-01-24
layers/meta-opentrons: add mosquitto

mjhuff:
Closes RET-1417

This PR adds support for Mosquitto, a message broker that implements MQTT.

Why Mosquitto?
- It's lightweight.
- It's open source.
- It's well-documented and supported.

See [this confluence doc](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3928621089/Notifications) for more background on the notifications work. [Here's the monorepo PR](https://github.com/Opentrons/opentrons/pull/14343). 

I'm fairly confident that the config settings are what we want, but feel free to give them a lookover [here](https://mosquitto.org/man/mosquitto-conf-5.html) if you're interested. Although we use default config values, we do need to explicitly specify these values. Otherwise, mosquitto rejects any connection attempt that isn't localhost. I didn't touch any setting related to logging/log rotation, so let me know if that needs to be done.


sfoster1 (approved):
Looks good to me!

vegano1 (approved):
Great work

PR #126, opened 2024-01-26
layers/meta-opentrons: fix app service unit

sfoster1:
This service unit has a bad dep! And while the app still starts, weirdly enough, it also breaks devtools.


vegano1 (approved):
nice, this was the last one

PR #127, opened 2024-01-26
layers/meta-opentrons: fix jupyter install in kirkstone

sfoster1:
For some reason, meta-jupyter on kirkstone has a weird package architecture that does not properly represent jupyter's dependencies. They seem to want you to install a packagegroup instead of the jupyter meta-package, since the meta-package now doesn't actually root a tree of all jupyter's deps. The problem is that (a) the package group wants pydot which needs graphviz from meta-openembedded which seems straight up broken and (b) even if you skip that it makes the image fail for reasons I don't fully understand.

Instead, we can just iterate on adding deps exclusively to our recipe that also depends on the jupyter metapackage until the thing can actually run notebooks, and fix up some other recipe and unit issues while we're at it.

This pr is successful if a build lets you run python commands in a notebook, and also pop a terminal and run terminal commands.


Laura-Danielle (approved):
exciting stuff

vegano1 (approved):

PR #128, opened 2024-01-26
layers/meta-opentrons: nodejs 18.18.2

sfoster1:
Backport nodejs 18.18.2 from mickledore and prefer it in the distro


vegano1 (approved):
heck yeah,
the future is now.

PR #131, opened 2024-02-02
layers/meta-opentrons: add boot script runner

sfoster1:
This is on the OT-2 and we forgot about it up to now - provide a way for people to run stuff on the system if they want. They have to install manually, but by dropping executable scripts in
/var/user-packages/etc/boot.d/ the scripts should be run at boot.


strangemonad:
ooo thank you thank you @sfoster1. Just upgraded our fleet. We were missing this badly

sfoster1:
> ooo thank you thank you @sfoster1. Just upgraded our fleet. We were missing this badly

Glad it helped! And sorry we missed it.

PR #133, opened 2024-02-06
layers/meta-opentrons: disable electron gpu usage

sfoster1:
This breaks when using SVG animations.

Closes RQA-2287
Closes RABR-22


vegano1 (approved):
nice,
Does this slow down animations at all or have any adverse effects on performance?

sfoster1:
Doesn't seem to!

PR #135, opened 2024-02-14
layers/meta-opentrons: address robot app g2d crashes

sfoster1:
It seems like occasionally weston runs out of space to get render buffers to allocate. This crashes weston and therefore crashes the ODD and therefore displays something weird and ugly because opentrons-loading can't run either.

Let's do two things to address this:
1. The ODD can autorestart, which should prevent the problem from lingering
2. We can increase the amount of contiguous memory available to ideally prevent it from happening

More info about contiguous memory on imx/galcore: https://community.nxp.com/t5/i-MX-Processors/imx8mp-How-to-resolve-weston-crash-with-g2d-alloc-alloc-memory/m-p/1535838

Closes RQA-2353


vegano1 (approved):
interesting, 
I thought we had disabled the GPU for rendering wayland

PR #136, opened 2024-03-06
layers/meta-opentrons: do not install gpsd

sfoster1:
It collides with a temperature sensor that we use in ABR.

Closes RABR-47

### DIFF
--- a/layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
+++ b/layers/meta-opentrons/recipes-images/images/opentrons-ot3-image.bb
@@ -59,6 +59,9 @@ IMAGE_INSTALL += " \
     opentrons-jupyter-notebook \
  "
 
+# gpsd is installed by packagegroup-tdx-cli for some reason and it breaks some integrations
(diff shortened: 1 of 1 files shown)


sfoster1:
Picks up the sensor without having to do anything to gpsd (because it is no longer there)
```
❯ ssh 10.14.11.148
root@Flex1234567890:~# cd /opt/opentrons-robot-server/
root@Flex1234567890:/opt/opentrons-robot-server# python3
Python 3.10.9 (main, Dec  6 2022, 18:44:57) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from hardware_testing.drivers import asair_sensor
>>> sensor = asair_sensor.BuildAsairSensor(False, True)

##########################################
#   Connecting to Environmental sensor   #
##########################################

Trying to connect to env sensor on port /dev/ttyUSB0
Found env sensor 008200ed on port /dev/ttyUSB0
>>> sensor.get_reading()
Reading(temperature=23.8, relative_humidity=12.7)
>>> 
```

SyntaxColoring (approved):

PR #142, opened 2024-04-04
fix: correct devtools and fix app scripts

sfoster1:
We tried to get nice readable scripts and environments in the app runscript and systemd unit and it didn't work and it's because of the sed thing and that's stupid so just write these values down in the files and check them in.

Also fix the devtools allow-origin to be pluralized so it actually works.
<https://opentrons.atlassian.net/browse/AUTH-285>


layers/meta-opentrons/recipes-robot/opentrons-robot-app/files/opentrons-robot-app.sh
```diff
@@ -0,0 +1,25 @@
...
+    fi
+fi
+
+# wait for weston
+while [ ! -e  $XDG_RUNTIME_DIR/wayland-0 ] ; do sleep 0.1; done
+sleep 1
+
+/opt/opentrons-app/opentrons \
+    --disable-gpu \
+    --remote-allow-origins=* \
```
sfoster1, line 16: this specifically is what was breaking devtools, this wasn't pluralized

koji: dev tools worked as expected.

vegano1 (approved):
Sweet,
This makes sense to me,
Thank you.

PR #143, opened 2024-04-04
layers/meta-opentrons: Update opentrons loading script to handle oem mode.

vegano1:
# Overview

We need to change the splash screen in user space, this is managed by the systems service `opentrons-loading. service`. The service uses the `opentrons-loading.sh` script to play the opentrons splash screen video using GStreamer CLI. This pull request expands on this script, it now handles OEM Mode and changes the splash screen accordingly.

The script reads the new `/var/lib/opentrons-system-server/system.env` file, this file is used by the system server to load and override system variables using dotenv. The system server updates the `OT_SYSTEM_SERVER_oem_mode_enabled` and `OT_SYSTEM_SERVER_oem_mode_splash_custom` via corresponding endpoints. These settings are then saved to the `system.env` file.

The `opentrons-loading.sh` script now looks for this `system.env` file, and if it finds it exports the variables so they can be used within the script. Once we have the variables we can determine if OEM Mode is enabled, and validate the location of the new custom OEM Splash screen if provided, otherwise we default to the given `oem_mode_default.png` file (The image is not final).

Closes: [PLAT-187](https://opentrons.atlassian.net/browse/PLAT-187)

# Test Plan

- [x] Create a build with the changes and make sure we can change the splash screen by changing the `system.env` file
- [x] Make sure the opentrons loading video plays if the `system.env` file is not found
- [x] Make sure the opentrons loading video plays if oem mode is NOT enabled
- [x] Make sure we set the default OEM mode splash screen if OEM mode is enabled but there is no custom OEM image
- [x] Make sure we set the custom splash screen if OEM Mode is enabled AND there is a custom OEM PNG file
- [x] Make sure we only accept PNG files for now, if some other type is provided play the opentrons loading video.
- [x] Make sure the robot-app service starts up after the loading service 
- [x] Make sure that the robot-app starts up even of the opentrons-loading.sh script fails and exits with exit code 0.

# Changelog

- Updated the `opentrons-loading.service` systemd file so we no longer pass in the resource we want to render
- Updated the `opentrons-loading.sh` script used by the `opentrons-loading.service` systemd service to take into account env file to change the splash screen according to OEM Mode. See the overview for a longer explanation.

# Review Requests

# Risk assessment

low, changes are not critical but will affect existing customers.

[PLAT-187]: https://opentrons.atlassian.net/browse/PLAT-187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


layers/meta-opentrons/recipes-robot/opentrons-robot-app/files/opentrons-loading.sh
```diff
@@ -1,16 +1,54 @@
...
+	else
+		echo "Default OEM file is set: ${oem_mode_splash_default}"
+		splash_screen_path=$oem_mode_splash_default
+	fi
+fi
+
+# Make sure the file exists
+if [ ! -f $splash_screen_path ]; then
+	echo "ERROR: Splash screen file not found: ${splash_screen_path}"
+	exit 1
```
sfoster1, line 40: this will make systemd consider this unit failed which might impact the app starting at all. may want to note the problem and have a fallback, or in any case exit 0 so it's still a success

sfoster1 (approved):
Looks good but I think we probably want this to exit 0 unconditionally so it doesn't interfere with further systemd unit startup

layers/meta-opentrons/recipes-robot/opentrons-robot-app/files/oem_mode_default.png
brenthagen, line 1: yes!!

PR #146, opened 2024-04-10
layers/meta-opentrons: various fixes for system server oem mode

vegano1:
A few fixes and additions for OEM Mode

- add `OT_SYSTEM_SERVER_dot_env_path` dotenv file to the systemd file for the system server
- use the linux `tr` command to strip away quotes and double quotes when parsing dotenv file
- loosen the regex when parsing `oem_mode_enabled` flag


sfoster1 (approved):
Nice

PR #150, opened 2024-05-15
chore(build): add concurrency to the github workflow to conditionally queue builds.

vegano1:
We make a lot of builds, which is great for having a pulse on the system. However, this leads to excessive builds that are never used, costing us computation time and money. This pull request fixes this problem (or greatly mitigates it) by adding Github [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency) to the workflow, which cancels queued builds with the same signature. The build signature is just a concatenated string of the monorepo, oe-core, and ot3-firmware refs, plus build variant used when building the image. Here are some examples,
```
'refs/heads/edge refs/heads/main refs/heads/main release'
'refs/tags/v7.3.0-alpha.2 refs/tags/v0.6.4 refs/tags/v50 release'
```
This means that a build will start right away if there are no other builds in progress; when another build with the same signature is added, it will be queued as they do now. However, if we now have a build in-progress, a build queued, and a new build with the same signature is added to the queue, the currently queued build will be canceled, as seen in the image below. So at any given time, we can have one in-progress build and, at most, one queued build of the same signature.

Screen Shot 2024-05-15 at 9 24 44 AM


sfoster1 (approved):
this is so blessed. thank you!

PR #154, opened 2024-07-01
layers/meta-opentrons: binding for opentrons cli

sfoster1:
Add explicit executable shell scripts for `opentrons_simulate` and `opentrons_execute`. 

Normally, when you install python packages that have script entrypoints, the install process will generate the scripts and put them in the appropriate place. Here, we have a separate python environment in a special directory that the system doesn't know about - it's used by setting environment variables in systemd units. That means the installation won't work. So we do it ourself here.

Closes RESC-285

### DIFF
--- a/layers/meta-opentrons/recipes-robot/opentrons-user-environment/files/opentrons_simulate
+++ b/layers/meta-opentrons/recipes-robot/opentrons-user-environment/files/opentrons_simulate
@@ -1,0 +1,5 @@
+#!/usr/bin/env sh
+export RUNNING_ON_VERDIN=1
+export OT_API_FF_enableOT3HardwareController="true"
+export PYTHONPATH=$PYTHONPATH:/opt/opentrons-robot-server
+python3 -m opentrons.simulate "$@"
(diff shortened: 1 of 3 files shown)


layers/meta-opentrons/recipes-robot/opentrons-user-environment/files/opentrons_simulate
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+export RUNNING_ON_VERDIN=1
+export OT_API_FF_enableOT3HardwareController="true"
```
vegano1, line 3: should this be false for simulate?

y3rsh:
Can you explain how this fixes the issue and then in what context these 2 files will be run?
Why are we fixing it here and not in the package?

ryanthecoder (approved):
looks good to me

layers/meta-opentrons/recipes-robot/opentrons-user-environment/files/opentrons_simulate
sfoster1: No, this is always true - this put it into "flex mode", it's not about controller vs simulator but about ot2 controller vs flex controller

sfoster1:
> Can you explain how this fixes the issue and then in what context these 2 files will be run? Why are we fixing it here and not in the package?

updated the pr description

PR #158, opened 2024-07-29
fix(CI): document consequences of tag refs

y3rsh:
## Overview

<https://opentrons.atlassian.net/browse/RDEVOPS-133>

- document that using tags as the ref in the workflow dispatch is for release only

### Edit - Keep all the workflow triggers the same - just document better the tag

> Changing this to handle RP events is more than I want to do right now.  Keep building on all pushes to `oe-core` 😊

### DIFF
--- a/.github/workflows/build-ot3-actions.yml
+++ b/.github/workflows/build-ot3-actions.yml
@@ -15,7 +15,7 @@ on:
         default: '-'
       ot3-firmware-ref:
         description: |
-          Ref of https://github.com/opentrons/ot3-firmware to build. ONLY use a tag ref if you are trying to actually release!!! It MUST be a full ref, e.g. refs/heads/main or  or '-' to indicate not-specified. If not specified, will be decided based on the monorepo ref; if that isn't specified, will be main.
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
Okay. So the new method is to start all builds manually when you're testing stuff?

.github/workflows/build-ot3-actions.yml
```diff
@@ -5,30 +5,33 @@ on:
...
+          What infra stage to run on. This should almost always be prod; dev is useful when you explicitly want to test or prod is busy.
         required: true
         type: choice
         options:
           - 'stage-prod'
-          - 'stage-staging'
           - 'stage-dev'
         default: 'stage-prod'
 
+concurrency:
```
vegano1, line 31: We have concurrency here

https://github.com/Opentrons/oe-core/blob/8aa4de42a68dde3a9630cdc2bb0ef4d4d3928ea4/.github/workflows/build-ot3-actions.yml#L67

y3rsh: Just saw that, fixing

y3rsh:
> Okay. So the new method is to start all builds manually when you're testing stuff?

No, I changed it back - triggering from PRs takes more work than I want to do right now.  Edited the PR and the Jira

sfoster1 (approved):
Nice!

vegano1 (approved):
lgtm

PR #160, opened 2024-08-07
opentrons-module-firmware: set the correct filename for the temperature module firmware.

vegano1:
The Flex/OT-2 robots' firmware update mechanism expects the temperature module firmware file to have the following format `temperature-module@v<version>.hex` ex. `temperature-module@v2.1.0`, however, we were saving the file as `temp-deck-arduino.ino.hex` which is incorrect. This pull request fixes this issue.


sfoster1 (approved):
Oh that's super annoying, thanks!

PR #163, opened 2024-08-20
Fix mlan0 crash

sfoster1:
Sometimes, the wifi adapter firmware crashes, and the device goes down. To fix (well, "fix") this, we can add a watcher script for NetworkManager's [dispatcher](https://man.archlinux.org/man/NetworkManager-dispatcher.8.en) interface, look for the down event, and [restart the bus](https://community.toradex.com/t/wifi-interface-not-appearing-on-verdin-imx8m-mini-after-resuming-from-suspension/22216/9?u=brayan.almonte) .


vegano1 (approved):
This should work around the issue, thank you.

sfoster1:
Put a build on a robot, ssh'd in, unbound the wifi interface, and it came right back!

PR #164, opened 2024-08-26
layers/meta-opentrons: overeager mlan0 reboot

sfoster1:
That was not how test works, and anyway we can just call it with normal pipe chaining.

This script would fire on any mlan0 down event, which could really mess up future interactions. Specifically, this would make the HTTP requests we use to connect to a wifi network fail, and then the robot would be connected after the interface got rebooted. Gross. Now it connects the first time.

Closes RQA-3087


vegano1 (approved):
Wow that's quite the bug, nice catch. Mlan0 should always be available even when the network is down, unless we are in the error state.

PR #166, opened 2024-09-05
meta-opentrons/systemd: add DNS + NTP servers that are reachable from China.

vegano1:
# Overview

The Flex cannot reach the Google DNS + NTP servers in the USA from within China, which causes unresolved domains and out-of-sync system time, leading to issues. So let's add a few backup DNS + NTP servers that are reachable from China so we don't run into this issue.

Closes: [PLAT-473](https://opentrons.atlassian.net/browse/PLAT-473)

# Changelog

- Add a custom list of DNS + NTP servers including ones reachable from China.

# Testing

- [x] systemd-timesyncd can sync the current datetime from one of the NTP servers
- [x] systemd-resolved can resolve domains from one of the DNS servers

[PLAT-473]: https://opentrons.atlassian.net/browse/PLAT-473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/layers/meta-opentrons/recipes-core/systemd/files/dns-fallback.conf
+++ b/layers/meta-opentrons/recipes-core/systemd/files/dns-fallback.conf
@@ -1,0 +1,5 @@
+# Google DNS is not available in China so we alternative.
+
+[Resolve]
+DNS=8.8.8.8 2001:da8::666
+FallbackDNS=1.1.1.1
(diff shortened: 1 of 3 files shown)


vegano1:
@sfoster1 
I changed to main branch, not critical for 8.0.

sfoster1:
If you rebase to latest main those extra commits will go away, but I don't think it matters anyway now that the actual code changes are in `main`

sfoster1 (approved):
Cool, let's do it! Probably rebase that last commit on main so the extra changes go away

PR #176, opened 2024-10-30
c-ares: change source branch from master to main

vegano1:
The [c-ares](https://github.com/c-ares/c-ares) repo changed its source branch from master to main, this pr fixes that on our end.


y3rsh (approved):
Awesome ty, I will cherypick this into the chore_release also.

PR #184, opened 2024-12-16
layers/meta-opentrons: add flex-stacker module to udev rules

ahiuchingau:
Let's identify the FLEX Stacker as an Opentrons module and give it a symlink!


vegano1 (approved):
Oh yeah!

PR #185, opened 2024-12-17
layers/meta-opentrons: support pydantic v2

sfoster1:
Contains the work to support pydantic v2 as in https://github.com/Opentrons/opentrons/pull/14871.

This is a pretty significant amount because pydantic v2 uses rust and modern python packaging tooling, which our older openembedded does not. So we're borrowing a lot from upstream openembedded-core and meta-openembedded to get a working rust cross setup, and hacking it together a little to support our pipenv installation process.


sfoster1:
We should squash merge this since it doesn't involve git shenanigans

vegano1 (approved):
super big rust dep! but can't be helped

PR #186, opened 2025-01-10
opentrons-systemd-units: add a systemd service to clear the fontconfig cache.

vegano1:
# Overview

The ODD did not recognize the simplified Chinese font when the system was updated from 8.2 or lower to 8.3.0, seems like something with Fontconfig caching the fonts from the previous installation was causing a problem. Clearing the Fontconfig cache fixes the issue, so let's add a `oneshot` systemd service that runs on startup and deletes all fontconfig cache in `/var/cache/fontconfig` and then rebuilds the fontcache with `fc-cache -rf` .

Closes: [RQA-3815](https://opentrons.atlassian.net/browse/RQA-3815)

# Testing

- update from 8.2 or lower to this build and make sure that Chinese characters are displayed correctly on the ODD.

[RQA-3815]: https://opentrons.atlassian.net/browse/RQA-3815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/layers/meta-opentrons/recipes-robot/opentrons-systemd-units/files/opentrons-clear-fontconfig-cache.service
+++ b/layers/meta-opentrons/recipes-robot/opentrons-systemd-units/files/opentrons-clear-fontconfig-cache.service
@@ -1,0 +1,6 @@
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/ot-clear-fontconfig-cache
+
+[Install]
+WantedBy=basic.target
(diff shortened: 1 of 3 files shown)


SyntaxColoring (approved):
Sounds good if it works and if this isn't something that takes, like, minutes to rebuild every boot.

Weird that this would be required. If you have an idea of why this is the case, it might be worth an explanatory comment in the script or the service file.

vegano1:
> Sounds good if it works and if this isn't something that takes, like, minutes to rebuild every boot.
> 
> Weird that this would be required. If you have an idea of why this is the case, it might be worth an explanatory comment in the script or the service file.

Naw it takes like 5 seconds to rebuild.

Unfortunately, I've not found out why this is happening,  but the standard practice seems to be to run fc-cache when a new font is installed. Just running that without removing everything with rm first does not fix the issue.

sfoster1 (approved):
Nice!

vegano1:
Successfully tested on Flex 8.2 -> 8.3 [build](https://github.com/Opentrons/oe-core/actions/runs/12716535151).

PR #189, opened 2025-02-05
python-byonoy: add updated absorbance 96 binary file to support future reference numbers.

vegano1:
This is the same V8 firmware version, which supports future devices with the reference number/SN  991-00335.


sfoster1 (approved):
ty!

PR #190, opened 2025-02-26
meta-opentrons/linux-kernel: use clone source for linux-stable-rt.

vegano1:
There is a problem accessing git.kernel.org, so let's use a clone source so we can fix the builds

```
fatal: unable to access 'https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/': Failed to connect to git.kernel.org port 443: Connection timed out 
```

Clone: https://kernel.googlesource.com/pub/scm/linux/kernel/git/rt/linux-stable-rt.git


sfoster1 (approved):
Ugh that's annoying

vegano1:
> Ugh that's annoying

I know :(

Link to build with this fix: https://github.com/Opentrons/oe-core/actions/runs/13553224868/job/37881632046

PR #191, opened 2025-02-27
meta-opentrons/linux-toradex-mainline: change LINUX_REPO source to https://kernel.googlesource.com with bbappend file.

vegano1:
# Overview

The previous [Opentrons/oe-core#190](https://github.com/Opentrons/oe-core/pull/190) PR did not fix the unreachable git.kernel.org source issue, unfortunately, this was hidden behind the fact that our [build](https://github.com/Opentrons/oe-core/actions/runs/13553224868) passed because the download/sstatecache/git were cached in S3. After manually cleaning the cache in S3 and EC2 action runner and re-running the build, it exposed the failed recipe, which this pull request should fix.

# Testing

- [x] [This](https://github.com/Opentrons/oe-core/actions/runs/13571807907/job/37938494129#logs) build should pass where they were failing before at the download sources step


sfoster1 (approved):
Looks good to me if the build passes.

y3rsh (approved):

vegano1:
Build has passed.

PR #195, opened 2025-04-14
recipes-robot/systemd: reduce the SyncIntervalSec from 5m to 1s to reduce memory usage.

vegano1:
This https://github.com/Opentrons/opentrons/issues/18034 bug identified some performance issues with the Flex robot, particularly with journal logging affecting the robot-server and robot-app processes. This pull request reduces the SyncIntervalSec from 5m to 1s to reduce the memory footprint from the logs before they are written to disk.

### DIFF
--- a/layers/meta-opentrons/recipes-robot/systemd/systemd-conf/opentrons-journald.conf
+++ b/layers/meta-opentrons/recipes-robot/systemd/systemd-conf/opentrons-journald.conf
@@ -1,4 +1,5 @@
 [Journal]
 Storage=persistent
 ForwardToSyslog=no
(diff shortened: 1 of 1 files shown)


SyntaxColoring (approved):
1 second feels very frequent to me. But that's just intuition. I don't have any reason to believe that this will be harmful, and I don't have any better suggestions.

Thank you!

sfoster1:
Yeah let's do it. It shouldn't be any more harmful than a writethrough.

vegano1:
Successfully tested on a Flex Devkit using the following build
https://github.com/Opentrons/oe-core/actions/runs/14450266229

PR #199, opened 2025-06-02
classes/pipenv_app_bundle: pin build deps

sfoster1:
A recent typing-extensions release seems to have broken pydantic-core's build on 3.10; pin that so it doesn't happen, and pin all the other build deps so it doesn't keep happening.


vegano1 (approved):
thank yo

PR #204, opened 2025-06-13
weston: increase galcore cma to fix g2d_alloc

sfoster1:
We were theoretically increasing the contiguous memory used by the galcore gpu driver, except that you can't use shell pipelines in systemd unit directives, so this was just printing the literal string "/bin/echo '671088640' > /sys/module/galcore/parameters/contiguousSize" to the unit file's standard out. Whoops!

Instead, use a little shell script that will echo the thing we want to the place we want. I don't know if there's a better way to do this.

Closes RQA-4285


vegano1 (approved):
might need the `increase-cma.sh` script

ddcc4:
> I don't know if there's a better way to do this.

Hi! I know on other distros, there's a file like `sysctl.d` to tweak parameters like these. Do you know if our OS has something like that, so we don't have to rely on shell scripting?

ddcc4 (approved):

PR #205, opened 2025-06-13
chore: add an autoupdate script

sfoster1:
The sources that we specify for the layers in openembedded are all taken from a google repotool manifest run by toradex. To update our openembedded, we have to update to the refs specified for all the layers by the repotool manifest. This is an annoying and manual pain.

Instead, here's a script that will just update all the submodules to a target ref of the repotool manifest. Run without --update to predict and with --update to execute. You'll have to commit the submodule changes if you want to afterwards.

This could theoretically be integrated into ci with some changes or a check mode that just prints if it would change anything pretty easily.


y3rsh (approved):
1. cloned fresh
2. `./update.sh`
3. `uv run set-refs-from-tdx-manifest.py 6.8.0 --update`
5. worked as expected setting the commit hash in the submodules.
6. `uv run set-refs-from-tdx-manifest.py 7.2.0 --update`
7. worked as expected setting the commit hash in the submodules.
8. tested with -vvv to see debug

Looks good.

set-refs-from-tdx-manifest.py
```diff
@@ -0,0 +1,534 @@
...
+    return manifest_path
+
+
+def get_manifest_repo(
+    manifest_target_path: Path, manifest_repo: str, quiet: bool
+) -> None:
+    """Get the manifest repo with git clone."""
+    LOG.info(f"cloning manifest {manifest_repo} to {manifest_target_path}")
+    git_results = git(
+        ["clone", manifest_repo, manifest_target_path],
```
vegano1, line 314: Maybe we can do a shallow clone? this takes a while to run

set-refs-from-tdx-manifest.py
```diff
@@ -0,0 +1,534 @@
...
+    path: str
+    url: str
+
+
+def _spec_from_lines(git_lines: str) -> Iterator[SubmoduleSpec]:
+    """Read a submodule spec from lines in .gitmodules"""
+    for git_line in git_lines:
+        if not git_line:
+            continue
+        flags = git_line[:2]
```
vegano1, line 64: what are submodule flags? also seems not used anywhere here?

set-refs-from-tdx-manifest.py
```diff
@@ -0,0 +1,534 @@
...
+        url = https://github.com/Freescale/meta-freescale.git
+
+    So you have to do a little stateful parser thing. This is pretty easy when you're tracking
+    the cursor implicitly by using python iterators, luckily. For parsing details see
+    parse_one_gitmodule; consider the gitmodules_lines iterator to be the cursor.
+
+    """
+    gitmodules_lines = iter(gitmodules_content.strip().split("\n"))
+    current_line = next(gitmodules_lines)
+    while True:
```
vegano1, line 149: very cool usage of iterators here!

vegano1 (approved):
This is super cool, thanks for putting this together!

- Could we move this script into the `scripts` dir
- Do we still need to `update.sh` script?

set-refs-from-tdx-manifest.py
sfoster1: they're little characters for like "modified since checkout". i guess i could drop the variable

set-refs-from-tdx-manifest.py
```diff
@@ -0,0 +1,534 @@
...
+        if manifest.path == repo_module.path:
+            yield (manifest, repo_module)
+            return
+    # if we don't get a path match, it might be because we put the repo in
+    # a different place than the manifest, so check in our overrides dict
+    if repo_module.path in _REPO_PATH_OVERRIDES:
+        for manifest in manifest_modules:
+            if manifest.path == _REPO_PATH_OVERRIDES[repo_module.path]:
+                yield (manifest, repo_module)
+                return
```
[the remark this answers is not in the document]
sfoster1: we could. i think it's clearer to keep the logic separate so comments can be interspersed, though.

set-refs-from-tdx-manifest.py
sfoster1: i love iterators and generator expressions. they're so good

PR #208, opened 2025-07-08
meta-opentrons/avrdude: use github instead of svn for avrdude.

vegano1:
# Overview

We cannot build avrdude because it is hosted on [http://svn.savannah.nongnu.org/](http://svn.savannah.nongnu.org/), but is unreachable, maybe because of [this](https://news.slashdot.org/story/25/07/06/1737253/the-fsf-faces-active-ongoing-and-increasing-ddos-attacks)? So let's migrate avrdude recipe from svn to github using the corresponding github [commit](https://github.com/avrdudes/avrdude/commit/f36484ed192369846e3af8a9c4cbc0352096ec84) hash from the one on svn `rev=1425`.

## Change log

- Use the alternative `avrdude` recipe hosted on GitHub instead of SVN


ddcc4 (approved):

sfoster1 (approved):
Oof. Do we need to do this for buildroot?

ddcc4 (approved):

PR #209, opened 2025-07-19
layers/meta-opentrons: Introduce opentrons-live-stream service including ffmpeg and nginx extensions

CaseyBatten:
## Overview
In order to support the usage of the embedded camera as well as third party cameras in the future FFMPEG has been introduced to give the most versatile toolset for camera/video interfacing. 

In addition to this the basic outline of the infrastructure for our live streaming service, called `opentrons-live-stream` is included in this PR. These modifications rely on the addition of the `nginx-rtmp-module` which allows our NGINX server to be configured with an RTMP live stream, delivered via FFMPEG. This also enables us to set up an HLS live stream that is browser friendly, which has been tested externally.

The structure of the opentrons-live-stream service in this PR is responsible for the creation of a .conf file that can be edited to change configurations about the live stream, such as the camera source we are sampling using FFMPEG, resolution of the stream, framerate and bitrate. In the future we may want to introduce other values to allow us to manipulate the camera feed locally rather than at client, such as video cropping and visual settings. IMPORTANT TO NOTE: These configurations are for the live stream only, not the the picture captured during a protocol run, though we may make them share settings if that is desired. 

### Stream Setup
The service will begin by defaulting the video source to NONE. In this case, the video stream will not begin on startup. To begin a video stream, edit the `SOURCE` variable of the opentrons-live-stream.conf file located at `/var/lib/opentrons-live-stream/` to be equal to whatever video source you desire in quotations. The embedded Flex camera defaults to `"/dev/video2"`. We currently support standard USB webcams so long as they appear as a linux video device as well.

After editing the configuration file, to start the stream restart the live stream service by executing the following:
`systemctl restart opentrons-live-stream`

The stream should appear at the following endpoints:
For RTMP (flv format, use a compatible player such as VLC of ffplay):
`rtmp://[ROBOT-IP]/live/stream`

For HLS (browser friendly, can be embedded in a web page player):
`http://[ROBOT-IP]:31950/hls/stream.m3u8`

### Notes on Stream Delay and Stream Type
The stream delay is shortest from source to client on the RTMP live feed, coming in at a minimum of 1-2 seconds when done though ffplay with the proper settings. However, to get a clear stream with limited frame drops and vsync issues the client player, such as VLC, will typically create a buffer and engage in some cleanup to ensure a good viewing experience. This brings the average delay to around 3 seconds.

The HLS stream works by sampling the stream on the robot side and creating a "playlist" for stream viewers to see on-demand. The sample size I have setup is currently 1 second, meaning the HLS stream will be 1 second slower than the RTMP stream at all times. However, not only is this format friendlier for browsers and embedded players, but we can adjust the playlist size as well to allow viewers to see footage that began prior to their connection to the live stream. The playlist is set to 3 seconds for now, so the last 3 "fragments" collected by the server are viewable after joining a stream. We could in theory increase the limit from 3 seconds up to multiple minutes, so that viewers "joining the stream late" could scroll back in time to see events that occurred prior to opening the stream on their player application. 

## Testing
This stream setup was tested on a robot in the lab. A 24/7 stream was left up for several days and did not negatively impact the performance of the device. FFMPEG CPU usage floated between 9-10% and NGINX CPU usage floated around 2-3% throughout.


layers/meta-opentrons/recipes-robot/nginx/files/nginx.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=The NGINX HTTP and reverse proxy server
+After=syslog.target network.target remote-fs.target nss-lookup.target
+
+[Service]
+Type=forking
+PIDFile=/run/nginx/nginx.pid
+ExecStartPre=/usr/sbin/nginx -t
+ExecStartPre=/bin/mkdir -p /var/www/localhost/html/stream
```
vegano1, line 9: do we need to run this before nginx bin commands?

PR #211, opened 2025-08-26
opentrons-module-firmware: stacker 1.0.1

sfoster1:
This version fixes an issue with post-update initialization.


vegano1 (approved):
Thank you

PR #214, opened 2025-09-08
refactor(meta-opentrons):  Add udev rule for system camera

mjhuff:
Closes [EXEC-1643](https://opentrons.atlassian.net/browse/EXEC-1643)

The USB camera that ships with the Flex seemingly gets assigned `/dev/video2` as its device node, but we'd like to rename the node to something more semantic, like `ot_system_camera`. Because this camera is not a module but is a device supported by the Flex, we'll need some sort of new naming convention here, and `ot_system` is the proposal. 

fad1045e9f931415abfbbf311dc374d6b48188e6 and [b82d256](https://github.com/Opentrons/oe-core/pull/214/commits/b82d25654134f45dc907451076ce6e7d025dd8cd) - Add the rule for the camera based on DEVPATH, utilizing the usb controller path. We have to include `ATTR{index}=="0"` as well, since our embedded camera offers a separate video node for camera metadata that does not support photo/video capture. 

6eb0bd6905d2f532b546c72212d8bbcc2e7d71d6 - Rename `95-opentrons-modules` to `95-opentrons-udev`, since this name more reflects how we use this file. I think it makes sense to keep all udev rules for modules and system devices in one place, but feel free to let me know if you disagree.

## Test Plan
Tested using [this build](https://opentrons.slack.com/archives/C81CR4VCZ/p1757518687055999), which includes the head of [this PR](https://github.com/Opentrons/opentrons/pull/19482) in the monorepo.

[EXEC-1643]: https://opentrons.atlassian.net/browse/EXEC-1643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfoster1 (commented):
This looks like it will work, but can we try identifying via device path (`DEVPATH` to udev) which will look like `/devices/host-controller-platform/bus/path`, since this is a built-in device that can be best uniquely identified by how it's plugged in

PR #217, opened 2025-09-15
fix(meta-opentrons): do not package x64 sentry cli binary

mjhuff:
Related to https://github.com/Opentrons/opentrons/pull/19569.

When we add the sentry vite config plugin, which uploads source maps to sentry on build, the packaging step's `objcopy` attempts to process `/node_modules/@sentry/cli-linux-x64/bin/sentry-cli`, which of course fails because we aren't on x64 architecture. To fix, let's just remove the x64 Sentry CLI binary before the packaging step. We don't _have_ to include it in the final build anyway, because source maps are created during the frontend build step, not by the CLI at runtime.

TO UNDRAFT: Does this work?


sfoster1 (approved):
Yeah, makes sense.

PR #220, opened 2025-09-26
layers/meta-opentrons: bump tc and hs firmware

sfoster1:
1.1.0 incorporates the error state querying changes.


sfoster1:
https://builds.opentrons.com/ot3-oe/18046422057/ot3-system.zip image from https://github.com/Opentrons/oe-core/actions/runs/18046422057/job/51358185509 has the right thing

PR #222, opened 2025-10-01
ci: added semver to tagging

neo-jesse:
(ci:feat):added semver to tagging

Handles any of the supported tag prefixes (v*, internal@, ot3@)
Uses proper semantic versioning with the semver package
Returns the highest version among the tags it receives
Is called with pre-filtered tags based on the variant and latestTagPrefixFor function


sfoster1 (approved):
Looks good I think but please make a more verbose PR description and eventual commit!

PR #264, opened 2026-01-16
chore: support ODD using pnpm

sfoster1:
Support https://github.com/Opentrons/opentrons/pull/18571 

todo: the sentry binary doesn't live in `node_modules` anymore it's in pnpm's symlink hell somewhere now


koji (approved):
thanks!

PR #270, opened 2026-01-29
feat(ci, runners): added Ephemeral runners to workflow

neo-jesse:
### Integrate ephemeral runner support into main build workflow

Integrates ephemeral runner functionality into build-ot3-actions.yml to address disk space issues on persistent self-hosted runners.

Changes:
* Adds request-ephemeral-runner job that requests a unique ephemeral runner for each workflow run
* Updates run-build job to use the ephemeral runner label, ensuring builds run on clean, isolated runners
* Fixes cache zip path construction to avoid realpath issues
* Adds S3 cache pull/push support for ephemeral runners with 24h TTL and collision-free naming

Benefits:
* Each build runs on a fresh runner, eliminating state-related issues
* Maintains cache performance via S3-backed shared cache for ephemeral runners


sfoster1 (dismissed):
let's do the python swap that we did in buildroot here too

sfoster1 (approved):
yay!

PR #298, opened 2026-03-24
fix(ci): use variable for SLACK_WEBHOOK_URL_TAGGED instead of secret

neo-jesse:
Read tagged Slack webhook URL from vars instead of secrets so it can be managed as a repository variable.


sfoster1 (approved):
looks good to me!

PR #301, opened 2026-04-08
git: switch meta-yocto url

sfoster1:
Per yocto maintainers, yoctoproject.org is now behind cloudflare to avoid scrapers and git urls must change:
https://lists.yoctoproject.org/g/yocto-infrastructure/message/227

This should fix recent build failures.


y3rsh (approved):
Thanks for catching!

PR #302, opened 2026-04-09
layers/meta-opentrons: https support in nginx

sfoster1:
We now use nginx for TLS termination using certificates provided by the key server.

This is a little annoying because nginx considers the presence of configured keys and certificates to be part of its config; if they're missing, the config is invalid, and if the config is invalid, nginx won't start. That means we can't use the same process for both our normal reverse proxy and the TLS terminating reverse proxy; when the reverse proxy fails nothing is accessible, and we want that to be robust.

So, we split our nginx configs and nginx systemd units into an `-http` and `-https` version; they both run at the same time (this is fine since they're on different ports) and talk to the same servers (this is fine since those servers are used to many things connecting to them, that's the point of them) and all is well.

## Testing
- Note that if you stop -https then -http still works fine
- Also everything should start up, as long as you have the right python on there


layers/meta-opentrons/recipes-httpd/nginx/files/nginx-http.conf
```diff
@@ -11,16 +12,14 @@ http {
...
-        # container build
         listen [::]:31950;
         listen 0.0.0.0:31950;
 
         client_body_in_file_only off;
         client_body_buffer_size  128k;
         client_max_body_size     2M;
 
-        access_log syslog:server=unix:/dev/log combined;
+        access_log /var/log/nginx/access-http.log combined;
```
SyntaxColoring: I don't think we want file-based logging, do we? It'll get pretty big over time. Shouldn't we either send this to syslog/journalctl so it can be rotated out via our usual config, or disable it entirely and rely on uvicorn's logs?

layers/meta-opentrons/recipes-httpd/nginx/files/nginx-https.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=TLS termination and reverse proxying for HTTPS login
```
SyntaxColoring: HTTPS isn't just for logins, right? Regular requests (e.g. `POST /runs`) also need to be HTTPS to protect the secret access token.

layers/meta-opentrons/recipes-httpd/nginx/files/nginx-https.conf
SyntaxColoring, line 1: Is this intentionally not quite in sync with `nginx-http.conf`? e.g. it's missing the special handling for `/system/time`.

SyntaxColoring (changes requested):
Fundamentally LGTM, but a couple questions about the config.

layers/meta-opentrons/recipes-httpd/nginx/files/nginx-https.conf
sfoster1: it's intentionally not quite in sync in that it doesn't have rtmp handling and such, but it probably should have `/system/time` yes

layers/meta-opentrons/recipes-httpd/nginx/files/nginx-http.conf
sfoster1: yeah, i suppose so

layers/meta-opentrons/recipes-httpd/nginx/files/nginx-https.conf
```diff
@@ -0,0 +1,86 @@
...
+        listen 0.0.0.0:32313 ssl;
+
+        client_body_in_file_only off;
+        client_body_buffer_size  128k;
+        client_max_body_size     2M;
+
+        access_log syslog:server=unix:/dev/log combined;
+        error_log stderr;
+
+        keepalive_timeout   70;
```
SyntaxColoring, line 28: Do you intentionally have this `keepalive_timeout` in the HTTPS conf but not the HTTP conf?

SyntaxColoring (approved):
Yay! LGTM if it's been tested.

layers/meta-opentrons/recipes-httpd/nginx/files/nginx-https.conf
sfoster1: yeah, the way i understand it you want more keepalive in tls connections because new tls connections mean new handshakes

PR #305, opened 2026-04-21
layers/meta-opentrons: Configure opentrons-auth-server persistence directory

SyntaxColoring:
It looks like we (I) just forgot to configure this, so opentrons-auth-server was using a temp directory each boot, so settings like `accessControlEnabled` weren't persisting across boots.


TamarZanzouri (approved):
LGTM!

sfoster1 (approved):
oops!

PR #306, opened 2026-04-25
fix(ci): chown cache before using

neo-jesse:
This PR hardens OT3 GitHub Actions cache cleanup by fixing ownership before cache archive/removal steps.
In .github/workflows/build-ot3-actions.yml, it chowns ${LOCAL_CACHE:-./cache} to the runner user before zipping/deleting cache contents, and makes the failure-path “Remove poisoned cache” step bash-based and non-blocking (continue-on-error) to avoid workflow failures caused by root-owned cache files left by Docker.


sfoster1 (approved):
ok but please alter the PR description and commit message, you're chowning it not rming it

PR #307, opened 2026-04-27
opentrons-robot-app: fix pnpm setup problems

sfoster1:
Our stack requires js-package-testing to be set up now because when pnpm invokes electron-rebuild it does so on _all_ the node_modules including js-package-testing, even though it's not part of the pnpm workspaces.


neo-jesse (approved):
Working now

PR #309, opened 2026-04-28
fix(ci): permission issues with build caches

neo-jesse:
Fixes permission failures on self-hosted runners when gathering artifacts, clearing the workspace, and wiping cache after failed builds.

Changes
* Docker image build: Pass --build-arg host_uid, host_gid, and username from the runner so bind-mounted oe-core and cache are written as the same UID as the job user (matches ot3Image.sh and the Dockerfile intent). Avoids relying on UID 1001 matching the runner (e.g. ec2-user as 1000).
* Gather results: chown the oe-core/build tree to the runner before creating deploy/opentrons and copying artifacts, for any leftover root- or wrong-UID files.
* Remove build data: Remove workspace children with sudo so root- or foreign-UID trees from Docker can still be deleted.
* Remove poisoned cache: After chown, run cache deletion with sudo so stubborn cache entries still clear.


.github/workflows/build-ot3-actions.yml
```diff
@@ -493,6 +500,10 @@ jobs:
         run: |
           _oe_build=$(pwd)/oe-core/build
           echo "found build dir at ${_oe_build}"
+          # Docker/BitBake may leave root-owned files in the bind-mounted build tree.
```
sfoster1, line 495: i think we do not want to do this because the oe-core runner wants to control these things. we can sudo when we delete stuff but probably not here

sfoster1 (approved):
thank you!

PR #310, opened 2026-05-08
layers/meta-opentrons: Hardware subprocess systemd service implementation

CaseyBatten:
Covers EXEC-2416

This is the implementation for the OT3API systemd service. For now we validate that its safe to boot by checking the feature flags and seeing if subprocess mode is enabled. This means that upon changing the feature flags we'll need a reboot for this version to start in "subprocess mode" officially. That would be to ensure that every process has the expected "resources" on the nameserver.

In a future PR I'd like to link this up with some stubbed out work on the robot server that lets us live-swap between subprocess mode and "classic" thread mode using systemd sockets to enable the switching without a reboot.

Question that only somewhat pertains to this PR, what do we want to do with the robot server if this service needs to restart for any reason? At that point, the robot server will need to "reconnect" with the new pyro instance of this service, which is fine from the OT3APIs perspective, but maybe we should be reporting a system failure on the robot server at that point? The other option is to just quietly handle it on the robot server side, which wouldn't be a problem between runs but during a run would be bad.

TESTS:
- [x] Test on an OT3 with subprocesses enabled, ensure this service is up and running on boot and that the robot server successfully finds this service and we can start running protocols
- [x] Test on an OT3 with subprocesses disabled, ensure we don't accidentally spin up a competing OT3API!


layers/meta-opentrons/recipes-robot/opentrons-robot-server/files/opentrons-hardware-api.service
```diff
@@ -0,0 +1,24 @@
...
+# If ExecConditions were met proceed to start the Hardware API subprocess
+ExecStart=python3 -m opentrons.hardware_control.pyro_utils.pyro_process_entry
+Environment=OT_API_FF_enableOT3HardwareController=true
+Environment=RUNNING_ON_VERDIN=true
+Environment=PYTHONPATH=/opt/opentrons-robot-server:/usr/lib/python3.10/site-packages
+Restart=on-failure
+TimeoutStartSec=10min
+
+[Install]
+WantedBy=multi-user.target
```
sfoster1: probably `WantedBy=opentrons-robot-server.service`, right?

layers/meta-opentrons/recipes-robot/opentrons-robot-server/files/validate-feature-flags.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+# Flags to validate
+HARDWARE_SUBPROCESS=$(jq -r '.enableHardwareSubprocess' /data/feature_flags.json)
```
sfoster1: can we inline this call into the unit? like `ExecCondition=/usr/bin/jq -r -e '.enableHardwareSubprocess?' /data/feature_flags.json`

i like separating things so it's less annoying to read, but here it's nice to have everything in one file

layers/meta-opentrons/recipes-robot/opentrons-robot-server/files/opentrons-hardware-api.service
```diff
@@ -0,0 +1,24 @@
+[Unit]
+Description=Opentrons Hardware API 
+
+Wants=nginx.service
```
sfoster1: i don't think we want the nginx dependencies. this doesn't interact with nginx in any way, right?

sfoster1 (changes requested):
couple little nits. in particular i don't think we want the nginx dependencies, do we?

layers/meta-opentrons/recipes-robot/opentrons-robot-server/files/opentrons-hardware-api.service
CaseyBatten: No I don't think so, you're right.

sfoster1 (approved):
looks good to me!

PR #311, opened 2026-05-12
recipes-robot/systemd: Reduce eMMC wear with optimized journald and fstab settings

vegano1:
# Overview

While part of the problem is on the Toradex end with the eMMC, there are some things we can do on our end, especially around logging. Right now, we write all logs to flash immediately and force-flush every second, which reduces the memory footprint of the logs. However, it also means we tear through the eMMC lifetime quickly. So let's reduce how often we write to flash when we sync journal logs and generally read/write to flash to reduce wear and tear.

Closes: [EXEC-2648](https://opentrons.atlassian.net/browse/EXEC-2648)

# Changelog

- Change the `SyncIntervalSec` in journal config from 1s to 30s to decrease emmc write rate by ~70%
- Added `commit=10` from 5 (default) to fstab to reduce filesystem journal writes
- Added `noatime` and `nodiratime` to fstab to eliminate unnecessary metadata reads/writes

# Testing and results

Created automated test scripts to systematically measure the **average write rate (MB/min)** and **RAM usage** across different configurations, with 5 minutes per config change.

### 1. SyncIntervalSec Test 

Here are the test results from changing the journal config `SyncIntervalSec=(1s 5s 10s 30s 1m)`
```
Interval   | Avg MB/min | Peak MB/2s | Avg RAM Used %
-------------------------------------------------------------
1s         |      11.11 |       1.37 |         58.4
5s         |       4.95 |       0.72 |         58.5
10s        |       4.19 |       0.57 |         58.5
30s        |       3.06 |       0.84 |         58.4
1m         |       2.75 |       0.63 |         58.5
```

### 2. Commit Test

Here are the test results from changing the fstab options `commit=(1s 5s 10s 30s 60s)`
```
commit=     | Avg MB/min | Peak MB/2s | Avg RAM Used %
--------------------------------------------------
1s         |      11.46 |       0.76 |         58.6
5s         |      10.33 |       0.68 |         58.7
10s        |      10.24 |       0.65 |         58.7
30s        |      10.91 |       0.91 |         58.8
60s        |      10.05 |       0.79 |         58.6
```

### 3. Combined 

Here are the test results from setting `SyncIntervalSec=10s` and changing the fstab `commit=(1s 5s 10s 30s 60s)`
```
commit=     | Avg MB/min | Peak MB/2s | Avg RAM Used %
--------------------------------------------------
1s         |       5.62 |       1.02 |         58.7
5s         |       4.19 |       1.91 |         58.6
10s        |       3.65 |       0.66 |         58.7
30s        |       3.29 |       0.67 |         58.7
60s        |       2.46 |       0.79 |         58.8
```

**Finding:** `SyncIntervalSec=30s` provides the best balance (~70% reduction in writes).
**Finding:** `commit=10s` is a good balance between frequent writes and not losing data on power loss. 
NOTE: We can probably increase this even further; we need to look into [Write-Ahead Logging](https://sqlite.org/wal.html) mode for SQLite to bypass the commit timeout.

# Reviewers

[EXEC-2648]: https://opentrons.atlassian.net/browse/EXEC-2648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/layers/meta-opentrons/recipes-core/base-files/files/custom-fstab
+++ b/layers/meta-opentrons/recipes-core/base-files/files/custom-fstab
@@ -14,5 +14,5 @@ tmpfs                /var/volatile        tmpfs      defaults              0  0
 /userfs/media            /media                none       defaults,bind,noatime,nodiratime,commit=10         0  0
 /userfs/var              /var                  none       defaults,bind,noatime,nodiratime,commit=10         0  0
 /userfs/var/mnt          /mnt                  none       defaults,bind,noatime,nodiratime,commit=10         0  0
-/userfs/etc/hostname     /etc/hostname         none       defaults,bind,nofail  0  0
-/userfs/etc/machine-info /etc/machine-info     none       defaults,bind,nofail  0  0
+/userfs/etc/hostname     /etc/hostname         none       defaults,bind,nofail,noatime,nodiratime,commit=10  0  0
+/userfs/etc/machine-info /etc/machine-info     none       defaults,bind,nofail,noatime,nodiratime,commit=10  0  0
--- a/layers/meta-opentrons/recipes-core/base-files/files/custom-fstab
+++ b/layers/meta-opentrons/recipes-core/base-files/files/custom-fstab
@@ -8,11 +8,11 @@ tmpfs                /var/volatile        tmpfs      defaults              0  0
 #/dev/mmcblk0p1       /media/card          auto       defaults,sync,noauto  0  0
 
 # opentrons custom mount points
-/dev/mmcblk0p4           /userfs               auto       rw,x-systemd.growfs   0  2
-/userfs/home             /home                 none       defaults,bind         0  0
-/userfs/data             /data                 none       defaults,bind         0  0
-/userfs/media            /media                none       defaults,bind         0  0
-/userfs/var              /var                  none       defaults,bind         0  0
-/userfs/var/mnt          /mnt                  none       defaults,bind         0  0
+/dev/mmcblk0p4           /userfs               auto       noatime,nodiratime,commit=10,rw,x-systemd.growfs   0  2
+/userfs/home             /home                 none       defaults,bind,noatime,nodiratime,commit=10         0  0
+/userfs/data             /data                 none       defaults,bind,noatime,nodiratime,commit=10         0  0
+/userfs/media            /media                none       defaults,bind,noatime,nodiratime,commit=10         0  0
+/userfs/var              /var                  none       defaults,bind,noatime,nodiratime,commit=10         0  0
+/userfs/var/mnt          /mnt                  none       defaults,bind,noatime,nodiratime,commit=10         0  0
 /userfs/etc/hostname     /etc/hostname         none       defaults,bind,nofail  0  0
 /userfs/etc/machine-info /etc/machine-info     none       defaults,bind,nofail  0  0
--- a/layers/meta-opentrons/recipes-robot/systemd/systemd-conf/opentrons-journald.conf
+++ b/layers/meta-opentrons/recipes-robot/systemd/systemd-conf/opentrons-journald.conf
@@ -1,8 +1,5 @@
 [Journal]
-Storage=volatile
+Storage=persistent
 ForwardToSyslog=no
-SyncIntervalSec=10m
-RuntimeMaxUse=64M
-RateLimitIntervalSec=30s
-RateLimitBurst=2500
+SyncIntervalSec=30s
 MaxLevelStore=debug
--- a/layers/meta-opentrons/recipes-robot/systemd/systemd-conf/opentrons-journald.conf
+++ b/layers/meta-opentrons/recipes-robot/systemd/systemd-conf/opentrons-journald.conf
@@ -1,5 +1,8 @@
 [Journal]
-Storage=persistent
+Storage=volatile
 ForwardToSyslog=no
-SyncIntervalSec=1s
+SyncIntervalSec=10m
+RuntimeMaxUse=64M
+RateLimitIntervalSec=30s
+RateLimitBurst=2500
 MaxLevelStore=debug


sfoster1 (approved):
Yeah, that's fine. We'll have to be okay with dropping logs. Good impetus to quiet everything down!

SyntaxColoring:
Wait, didn't we set it to 1s in response to a customer investigation that large syncs could stop the world? Wouldn't this regress on that?

[Edit: #195 is what I'm thinking of. That talks about memory usage, not large sync stalls, but I thought large syncs had something to do with the rationale.]

SyntaxColoring:
Also:

> This does mean that in the case of a power loss, we could lose up to 10 minutes' worth of logs

I'm worried that something is going to hang, and the user is going to power cycle their machine to break it out, and then we'll lose all of the debugging information. What about like 1m instead of 10m?

> Add RateLimitIntervalSec=30s and RateLimitBurst=2500 to limit the number of messages any given service can write

Have we tested these particular numbers? I suspect they're going to be hit very easily during a normal protocol run. I just ran `journalctl --since=30s` on a random office robot, and it returned ~1000 entries. And that's just with the ambient stuff like module polling. If it were actually moving the pipettes around....

layers/meta-opentrons/recipes-robot/systemd/systemd-conf/opentrons-journald.conf
```diff
@@ -1,5 +1,8 @@
 [Journal]
-Storage=persistent
+Storage=volatile
```
SyntaxColoring: According to [the documentation](https://www.freedesktop.org/software/systemd/man/latest/journald.conf.html#Storage=):

> If "volatile", journal log data will be stored only in memory

That's not what we want, is it?

SyntaxColoring (changes requested):

sfoster1:
> Also:
> 
> > This does mean that in the case of a power loss, we could lose up to 10 minutes' worth of logs
> 
> I'm worried that something is going to hang, and the user is going to power cycle their machine to break it out, and then we'll lose all of the debugging information. What about like 1m instead of 10m?

We could find some intermediate time, sure. 60x fewer writes is still pretty good.

> > Add RateLimitIntervalSec=30s and RateLimitBurst=2500 to limit the number of messages any given service can write
> 
> Have we tested these particular numbers? I suspect they're going to be hit very easily during a normal protocol run. I just ran `journalctl --since=30s` on a random office robot, and it returned ~1000 entries. And that's just with the ambient stuff like module polling. If it were actually moving the pipettes around....

We're definitely going to need to quiet down some of our logging to avoid this

layers/meta-opentrons/recipes-robot/systemd/systemd-conf/opentrons-journald.conf
vegano1: Yeah, we want to store in ram first, then sync periodically to flash.

SyntaxColoring: I don't think that is what `Storage=volatile` does, though. From my reading of the documentation, `Storage=volatile` *only stores to RAM,* and never syncs to persistent storage.

vegano1:
> Also:
> 
> > This does mean that in the case of a power loss, we could lose up to 10 minutes' worth of logs
> 
> I'm worried that something is going to hang, and the user is going to power cycle their machine to break it out, and then we'll lose all of the debugging information. What about like 1m instead of 10m?
> 
> > Add RateLimitIntervalSec=30s and RateLimitBurst=2500 to limit the number of messages any given service can write
> 
> Have we tested these particular numbers? I suspect they're going to be hit very easily during a normal protocol run. I just ran `journalctl --since=30s` on a random office robot, and it returned ~1000 entries. And that's just with the ambient stuff like module polling. If it were actually moving the pipettes around....

The system won't hang as it did before, because now we actually have limits on how much RAM is used for logs before we dump to flash; this was not the case before. This is done with `RuntimeMaxUse=64M`  which limits the RAM log usage before we sync with `SyncIntervalSec=10m`.

There is still the risk of losing minutes' worth of logs if they lose power; that's something we have to find a happy medium with losing logs vs degrading the SOM flash, which is a lot more critical and costly. I don't mind losing logs here and there if it means we can guarantee that the SOM flash will last the lifetime of the Flex.

We do want reasonable numbers for both how often we sync and rate limit, so ill be doing some testing to verify these values.

layers/meta-opentrons/recipes-robot/systemd/systemd-conf/opentrons-journald.conf
vegano1: That's what `SyncIntervalSec=10m` is for, I'm verifying this.

vegano1:
@SyntaxColoring @sfoster1 

I ended up removing the versatile and rate limiting changes, they did not behave as I'd expect. Instead we increased the SyncIntervalSec to 30s from 1s, which reduces read/writes by quite a lot (see findings).

I also updated how we mount the /users persistent directory via the fstab

Added noatime and nodiratime to fstab to eliminate unnecessary metadata reads/writes

Added commit=10 from 5 (default) to fstab to reduce filesystem journal writes

layers/meta-opentrons/recipes-core/base-files/files/custom-fstab
```diff
@@ -8,11 +8,11 @@ tmpfs                /var/volatile        tmpfs      defaults              0  0
...
-/userfs/var              /var                  none       defaults,bind         0  0
-/userfs/var/mnt          /mnt                  none       defaults,bind         0  0
+/dev/mmcblk0p4           /userfs               auto       noatime,nodiratime,commit=10,rw,x-systemd.growfs   0  2
+/userfs/home             /home                 none       defaults,bind,noatime,nodiratime,commit=10         0  0
+/userfs/data             /data                 none       defaults,bind,noatime,nodiratime,commit=10         0  0
+/userfs/media            /media                none       defaults,bind,noatime,nodiratime,commit=10         0  0
+/userfs/var              /var                  none       defaults,bind,noatime,nodiratime,commit=10         0  0
+/userfs/var/mnt          /mnt                  none       defaults,bind,noatime,nodiratime,commit=10         0  0
 /userfs/etc/hostname     /etc/hostname         none       defaults,bind,nofail  0  0
 /userfs/etc/machine-info /etc/machine-info     none       defaults,bind,nofail  0  0
```
SyntaxColoring: If we're adding these options to e.g. `/var`, should we also add them to `/etc/hostname` and `/etc/machine-info`? Both of those are probably read pretty frequently.

SyntaxColoring (approved):
Makes sense, thank you!

layers/meta-opentrons/recipes-core/base-files/files/custom-fstab
sfoster1: probably the other way around - i think we don't need these for the rest of the bind mounts, just line 11

sfoster1 (approved):
Looks good, but I'd prefer having the options just on the actual fs mount rather than the bind mounts

mjhuff:
Think we need to target 9.1.0 before merging?

layers/meta-opentrons/recipes-core/base-files/files/custom-fstab
vegano1: @sfoster1 

Access time flags like noatime and nodiratime only apply to the mount point, not the underlying physical drive. Since bind mounts are virtual filesystems, I don't think they inherit the options from the parent and have to be specified.

[Virtual Filesystem Notes](https://man7.org/linux/man-pages/man8/mount.8.html#:~:text=The%20options%20nosuid,VFS%20options%2C%20try%3A)

[per-mount basis](https://man7.org/linux/man-pages/man2/mount.2.html#:~:text=From%20Linux%202.4%20onward%2C%20some%20of%20the%20above%20flags%20are%20settable%20on%20a%0A%20%20%20%20%20%20%20per%2Dmount%20basis%2C%20while%20others%20apply%20to%20the%20superblock%20of%20the%0A%20%20%20%20%20%20%20mounted%20filesystem%2C%20meaning%20that%20all%20mounts%20of%20the%20same%20filesystem%0A%20%20%20%20%20%20%20share%20those%20flags.%20%20(Previously%2C%20all%20of%20the%20flags%20were%20per%2D%0A%20%20%20%20%20%20%20superblock.))

sfoster1 (approved):
I think if those apply to access points you do need to add them to `/etc/hostname` and `/etc/machine-info` as well. once you do that, good to merge!

PR #312, opened 2026-05-13
fix(pnpm): monorepo using pnpm vs yarn

neo-jesse:
Makes oe-core image builds work when the monorepo checkout still declares Yarn in packageManager (e.g. chore_release-pd-8.10.2 tags) while the robot-app recipe continues to use pnpm.

Changes
opentrons-robot-app.bb: Set pnpm cache via global config from / so pnpm config does not run in a Yarn-only workspace; strip packageManager from root package.json before pnpm install / pnpm exec; keep package-manager-strict=false for install, electron-rebuild, and electron-builder.


sfoster1 (changes requested):
wait we don't need to do this if it's for `release-pd`. that's a different application and doesn't need flex builds

PR #313, opened 2026-05-18
layers/meta-opentrons: Isolate temporary directories and clean them up when services exit

SyntaxColoring:
# Overview

This closes https://opentrons.atlassian.net/browse/EXEC-2673 and goes towards https://opentrons.atlassian.net/browse/EXEC-2672.

# Details

We currently have this funny failure mode:

1. A service starts.
2. The service consumes a bunch of memory.
3. The service writes a bunch of crap to `/tmp`, which consumes even more memory. (Our `/tmp` is pure RAM, a `tmpfs` mount, with no spillover to flash.)
4. The service gets killed because it's using too much memory. The _process's_ memory is released. **But all of the crap that it added to `/tmp` will be left behind, still consuming memory.**
5. systemd restarts the service. Go back to step 1.

Every iteration accumulates more crap in `/tmp`, consuming more memory. After a couple iterations, the system will be totally unresponsive, including basic interaction over SSH.

To fix that, this PR enables [PrivateTmp](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#PrivateTmp=) on all long-running Opentrons services. The main use of `PrivateTmp` is that it isolates services from each other, but for this fix, we're taking advantage of a side benefit: if a service fails, systemd can clean up its temp files immediately.

This will not fix the underlying resource exhaustion, but I'm hoping it it will keep the system more responsive and debuggable. The service will still get caught in a startup loop, but its memory usage should now reset to zero after each iteration.

# Test plan

* [x] Try the reproduction steps in EXEC-2672 and see if anything's improved.

# Review requests

I'm not aware of any place where we're deliberately using `/tmp` to share files across services, but double-check me.

# Risk analysis

Medium. This will isolate the services' `/tmp` directories from each other. If we were relying on `/tmp` for inter-service communication, then that will break with this PR. For example, if `opentrons-robot-app` is ever doing something like "hey `robot-server`, go import this file that I've placed in `/tmp/opentrons-robot-app`."

### DIFF
--- a/layers/meta-opentrons/recipes-robot/opentrons-auth-server/files/opentrons-auth-server.service
+++ b/layers/meta-opentrons/recipes-robot/opentrons-auth-server/files/opentrons-auth-server.service
@@ -12,6 +12,8 @@ Environment=PYTHONPATH=/opt/opentrons-auth-server
 Environment=OT_AUTH_SERVER_persistence_directory=/var/lib/opentrons-auth-server
 Restart=on-failure
 TimeoutStartSec=3min
+# We're using PrivateTmp for its feature of automatically cleaning up if the service crashes.
+PrivateTmp=yes
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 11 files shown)


vegano1 (approved):
This makes total sense, thank you!
Very important for services that we start/stop frequently, like the live stream service

PR #317, opened 2026-06-03
layers/meta-opentrons: Update heatershaker ref

ryanthecoder:
New heatershaker firmware fixes a race condition that results in false positives in the heater circuit open/shorted check.


sfoster1 (approved):
ty!

PR #326, opened 2026-06-16
update vacuum module fw version to v0.0.8

caila-marashaj:
## Overview
Bump the vacuum module fw version to 0.0.8 since the dvt firmware is tested in the factory and working


sfoster1 (approved):
sweet, ty

PR #329, opened 2026-06-17
build-refs: map external stack tags to ex* on ot3-firmware

y3rsh:
## Overview

External coordinated Flex releases use semver `v*` tags on `opentrons` and `oe-core`. Checking out the same semver tag on `ot3-firmware` breaks cmake `git describe --match=v*` (semver tags match the glob but not the integer version regex), which produced invalid firmware version metadata and failed builds.

This PR updates build-refs so `ot3-firmware` uses a separate coordination tag namespace:

- **External:** `v9.1.0-alpha.7` (stack) maps to `ex9.1.0-alpha.7` (ot3-firmware checkout)
- **Internal:** `ot3@8.5.0-alpha.1` is unchanged on ot3-firmware
- **Integer version tags** (`v70`) and branch builds are unchanged

Release engineers dual-tag ot3-firmware commits with the coordination tag plus an integer `vN` tag. CI checks out the coordination tag; cmake discovers the co-located `vN` for the firmware version integer.

Acceptance criteria:
- Tag dispatches for external releases resolve ot3-firmware to `ex*` tags
- Internal `ot3@` tag builds still resolve ot3-firmware to the same `ot3@` tag
- Branch/default builds are unchanged
- README documents the dual-tag release workflow

Related: fixes coordinated tag builds that failed when ot3-firmware was checked out at semver `v*` tags (e.g. `v9.1.0-alpha.7`).

## Why runtime behavior is unchanged

This PR only changes **which git ref CI checks out** on ot3-firmware for tag builds. It does not change cmake, manifest generation, or robot firmware update logic.

```mermaid
flowchart TD
  A["Stack dispatch tag v9.1.0-alpha.7 or ot3@8.5.0-alpha.1"] --> B["build-refs maps ot3-firmware ref ex* or ot3@"]
  B --> C["CI checks out coordination tag on ot3-firmware"]
  C --> D["CMake generate_version.cmake git describe match v*"]
  D --> E["Co-located vN tag on same commit e.g. v70"]
  E --> F["version.c embedded in MCU binaries"]
  C --> G["cmake install subsystem_versions.py same git describe rule"]
  G --> H["opentrons-firmware.json robot manifest"]
  H --> I["Robot compares installed shortsha vs manifest shortsha"]
  F --> I
```

Evidence from ot3-firmware (version integer is always from `vN`):

- [`common/core/generate_version.cmake`](https://github.com/Opentrons/ot3-firmware/blob/main/common/core/generate_version.cmake): `git describe --match=v*`, regex extracts integer `v([0-9]+)` into `version.c`
- [`common/core/version.c.in`](https://github.com/Opentrons/ot3-firmware/blob/main/common/core/version.c.in): baked-in `uint32_t version` reported by each subsystem over CAN
- [`scripts/subsystem_versions.py`](https://github.com/Opentrons/ot3-firmware/blob/main/scripts/subsystem_versions.py): same `git describe --match=v*` when writing `opentrons-firmware.json` at install time

Why this is clearer than resolving "latest" or aliasing to integer tags in build-refs:

- **No implicit tag discovery.** We do not list tags, sort by date, or pick a "latest" integer tag. The release engineer places `ex*`/`ot3@` + `vN` on the commit; build-refs only renames the external stack tag to `ex*`.
- **Matches existing cmake contract.** ot3-firmware already assumed integer `vN` tags for version embedding; coordination tags just must not start with `v` so `git describe --match=v*` is unambiguous.
- **Robot upgrade/downgrade unchanged.** Updates are gated by git **shortsha** mismatch in the installed manifest vs device, not by integer version comparison. Shipping an older `vN` with a new SHA still updates (or downgrades) the same way as before.

Branch/develop builds are also unchanged: build-refs still resolves ot3-firmware to matching branch names, and cmake continues to use the nearest reachable `v*` tag from HEAD.

## Test Plan and Hands on Testing

- [x] `cd .github/actions/build-refs && npm test` (54 tests pass)
- [x] `cd .github/actions/build-refs && npm run build` (dist rebuilt)
- [x] CI `test-build-refs` workflow on this PR

## Changelog

- build-refs maps external stack semver tags to `ex*` coordination tags on ot3-firmware
- Removed unused `latestTag` helper and semver dependency from build-refs action
- Documented ot3-firmware dual-tag release workflow (`ex*`/`ot3@` + `vN`) in README
- generate Summary text for the workflow right from the action 

**Breaking change (release process):** ot3-firmware external releases must use `ex*` coordination tags instead of semver `v*` tags. Integer `vN` tags remain required on the same commit for cmake version embedding.

## Risk assessment

- **Attention level:** medium (release ref resolution for ot3-firmware tag builds)
- **Blast radius:** Flex tag-dispatch builds only; branch/develop builds unaffected
- **Mitigation:** Unit tests for mapping, existing build-refs CI workflow, README documents required tags


.github/actions/build-refs/action/index.ts
```diff
@@ -38,8 +37,152 @@ export function restAPICompliantRef(input: Ref): string {
...
 
-export interface GitHubApiTag {
-  ref: Tag
+export function tagNameFromRef(ref: Tag): string {
+  return ref.slice('refs/tags/'.length)
+}
+
+/** Integer-only firmware version tags such as v70 (not stack semver v9.1.0). */
+export function isFirmwareVersionTagRef(ref: Ref): boolean {
+  if (!ref.startsWith('refs/tags/v')) {
```
sfoster1:
```suggestion
  if (!ref.startsWith('refs/tags/')) {
```

aligns with what `tagNameFromRef` requires, and the regex checks for the v anyway

.github/actions/build-refs/action/index.ts
```diff
@@ -293,14 +372,26 @@ async function run() {
 
   const attemptable = Array.from(inputs.entries()).reduce(
     (prev: AttemptableRefs, [repoName, inputRef]): AttemptableRefs => {
+      const normalizedInput =
+        repoName === FIRMWARE_REPO && inputRef
+          ? normalizeFirmwareInputRef(inputRef)
```
sfoster1: `normalizeFirmwareInputRef()` returns:
- the input, if the input isn't a prefixed tag (which should only happen through user error, right? since this is dealing with the firmware repo tag input?)
- a prefixed ex-tag, if this is a stack release tag, which also should only happen through user error, right? this is the firmware repo tag input, and the firmware repo should not have stack release tags
- null, if it's a prefixed tag that isn't a stack-release tag

so is the idea here that this is for removing a couple categories of input error? or is the idea that you _should_ be entering a stack tag into the input for ot3-firmware?

sfoster1: also having a ternary here _and_ a check for lack of prefix in `normalizeFirmwareInputRef` is duplicative

sfoster1 (approved):
this looks ok to me. the code is confusing and has a lot of duplicative checks for the presence of `refs/tags` in various places that all feed into each other. i'd prefer to see branded types for has-prefix vs has-no-prefix vs prefix-unknown used to cut down on that stuff.

.github/actions/build-refs/action/index.ts
y3rsh: Collapsed the duplicate helpers into firmwareRefForStackRef and routed all firmware tag resolution through it.

PR #333, opened 2026-06-22
fix(slack,cache): changed order of cache sync

neo-jesse:
fix(slack,cache): changed order of cache sync

Changed cache sync to AFTER the S3 deploy completion.

### DIFF
--- a/.github/workflows/build-ot3-actions.yml
+++ b/.github/workflows/build-ot3-actions.yml
@@ -586,6 +586,12 @@ jobs:
           artifacts: ${{ steps.artifacts.outputs.versioned_version_json }}
           artifactContentType: application/json
           token: ${{secrets.MONOREPO_RELEASE_ARTIFACT_UPLOAD_TOKEN}}
+      # After S3 artifact upload (and monorepo release uploads when applicable), notify
+      # Slack that build outputs are deployed—not that cache sync or cleanup finished.
+      # Conditions: prior steps succeeded; run was workflow_dispatch (manual or cross-repo
+      # dispatch from opentrons tag/branch push). Branch builds always alert. Tagged monorepo
+      # builds alert only when decide-refs resolved a variant (internal-release or release),
+      # skipping tags that do not map to a publishable Flex image variant.
       - name: Send build deployed alert
         if: success() && github.event_name == 'workflow_dispatch' && (!startsWith(needs.decide-refs.outputs.monorepo, 'refs/tags/') || needs.decide-refs.outputs.variant != '')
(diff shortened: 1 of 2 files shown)


.github/workflows/build-ot3-actions.yml
```diff
@@ -607,6 +586,59 @@ jobs:
...
+          workflow_name: "Build Flex image on github workflows"
+          webhook_url: ${{ secrets.SLACK_WEBHOOK_URL }}
+          console_url: ${{ steps.upload-results.outputs.console_url }}
+          system_url: ${{ steps.upload-results.outputs.system_url }}
+          fullimage_url: ${{ steps.upload-results.outputs.fullimage_url }}
+          version_file_url: ${{ steps.upload-results.outputs.version_file_url }}
+          release_notes_file_url: ${{ steps.upload-results.outputs.release_notes_file_url }}
+          oe-core_ref: ${{ needs.decide-refs.outputs.oe-core }}
+          monorepo_ref: ${{ needs.decide-refs.outputs.monorepo }}
+          firmware_ref: ${{ needs.decide-refs.outputs.ot3-firmware }}
```
y3rsh: Seems like instead of having a job for sending tagged and untagged we just have one job and tag becomes a part of the data in the alert. Set to None or the tag.

neo-jesse: Fixed.

.github/workflows/build-ot3-actions.yml
```diff
@@ -607,6 +586,43 @@ jobs:
           artifacts: ${{ steps.artifacts.outputs.versioned_version_json }}
           artifactContentType: application/json
           token: ${{secrets.MONOREPO_RELEASE_ARTIFACT_UPLOAD_TOKEN}}
+      - name: Send build deployed alert
+        if: success() && github.event_name == 'workflow_dispatch' && (!startsWith(needs.decide-refs.outputs.monorepo, 'refs/tags/') || needs.decide-refs.outputs.variant != '')
```
y3rsh: ~~Could we get a comment here that explains this conditional.  I would like the comment to explain the actual logic and the intention of this job.
Please add the same kind of comment to the failure job below.~~

~~I think I follow that the workflow dispatch runs are the ones we most care about because that is the event type when jobs are dispatched due to a monorepo tag or a Workflow dispatch manual run.  So we are only sending Notifications on those.~~

y3rsh:
I think take all changes to the slack message out of this PR, just change the location of the cache sync so that it happens after the asset upload so that releases happen faster.  I would like to re-write completely the slack action in another PR

y3rsh (approved):

PR #334, opened 2026-06-23
feat(layers/meta-opentrons): Add smartscan barcode scanner to udev rules

ryanthecoder:
This is the brand of scanner they want to use on the line so this just adds the udev rules, driver is added in a separate monorepo PR


caila-marashaj (approved):

sfoster1 (approved):
sure. at some point we may want to make these more general, having a specific PID seems a little tough to maintain

PR #336, opened 2026-06-25
scripts: push builds to productionV2 key

sfoster1:
This means that pre-9.1.1 ODD applications will no longer see and automatically download updates, preventing resource exhaustion issues from being triggered on robots running 9.0.0 whenever we do a new release.


mjhuff (approved):
Makes sense, thank you!

sfoster1:
Unfortunately we can only really test this by building an alpha.

PR #337, opened 2026-06-26
fix(meta-opentrons): add the cp210x driver back in

ryanthecoder:
The cp210x driver must have been removed from the default drivers when we updated linux at some point.

We need this driver for the force gauge used in hardware-testing.


sfoster1 (approved):
whoops!

PR #348, opened 2026-07-13
ci: STEP 1: ECR BuildKit cache for oe-core Docker image

neo-jesse:
## Summary
- Cache the CI `Dockerfile` build in ECR via BuildKit (`cache-from` / `cache-to`) when `OT3_OE_DOCKER_ECR_REPOSITORY_{DEV,PROD}` is set.
- Pin `aws-actions/amazon-ecr-login` to the peeled v2.0.1 commit SHA (zizmor-safe).
- Document the ECR cache variables and required IAM permissions.

Needs applied first: https://github.com/Opentrons/robot-stack-infra/pull/63

## Test plan
- [x] Stage-dev build with ECR repo vars set: Docker build reuses layers on a second run
- [x] Confirm zizmor passes on the workflow
- [x] Unset ECR vars still falls back to plain `docker build`


neo-jesse:
Remote Caching: For CI/CD environments with ephemeral runners, BuildKit supports external backends to persist cache between builds:

Registry: Push/pull cache layers to/from a container registry using --cache-from and --cache-to flags. 
GitHub Actions: Use type=gha for native integration with the GitHub cache API.
Local/Filesystem: Export to a shared directory for persistent storage on self-hosted runners. 
Cache Mounts: Beyond layer caching, BuildKit uses RUN --mount=type=cache to persist specific directories (e.g., package managers' caches) across builds. These mounts do not add layers to the final image, keeping images small while significantly speeding up dependency installation and compilation. 

Use mode=max with remote cache exports to include intermediate build stages in the cache, not just final layers.

neo-jesse:
Test run that compares the 2 Docker build type/images: https://github.com/Opentrons/oe-core/actions/runs/29526428162/job/87715886145

Verdict: functionally the same. Classic docker build and Buildx (docker-container driver, same as CI, without ECR) produce equivalent images.

### Docker image equivalence (`docker build` vs Buildx)

Ran classic `docker build` and Buildx (`docker-container` driver, same as CI, no ECR) side-by-side:
https://github.com/Opentrons/oe-core/actions/runs/29526428162

| Check | Result |
| --- | --- |
| Image size | Both **2153371958** bytes |
| File list (`docker export \| tar -t`) | **Identical** |
| Installed packages (`dpkg -l` hash) | **Identical** (`33c80eb2…`) |
| `uv` / Python / locale / uid | Both `uv 0.9.17`, Python 3.12.13, `LANG=en_US.UTF-8`, uid 1001 |
| Image ID / full export sha256 | **Differ** (expected — timestamps / builder metadata) |

.github/workflows/build-ot3-actions.yml
```diff
@@ -333,6 +335,25 @@ jobs:
...
+            echo "OT3_OE_DOCKER_ECR_REPOSITORY_* not set for ${{ matrix.build_env }}; Docker build will run without ECR layer cache."
+          else
+            echo "Using ECR repository ${ecr_repo} for BuildKit layer cache"
+          fi
+          echo "ECR_REPOSITORY=${ecr_repo}" >> "$GITHUB_ENV"
+      - name: Login to Amazon ECR
+        if: ${{ env.ECR_REPOSITORY != '' }}
+        uses: aws-actions/amazon-ecr-login@062b18b96a7aff071d4dc91bc00c4c1a7945b076 # v2.0.1
+        with:
+          mask-password: false
```
sfoster1: ?

sfoster1 (changes requested):
looks good but is that `mask-password: false` thing on purpose?

.github/workflows/build-ot3-actions.yml
neo-jesse: Looks like a migration leftover, set to true now

sfoster1 (approved):
nice, thank you!

PR #349, opened 2026-07-13
ci: STEP 2: BitBake S3 cache as tar.zst with fingerprint skip

neo-jesse:
## Summary
- Add `.github/scripts/s3-bitbake-cache.sh` for `downloads` / `sstate` / `git` as `tar.zst` (legacy zip fallback on pull).
- Push skips archive+upload when a content fingerprint matches the remote `*.manifest`.
- Harden pull/push error reporting and zstd install / zip fallback.
- Wire the build workflow to use the helper instead of inline zip logic.

## Test plan
- [x] Stage-dev warm build: pull finds caches, BitBake gets high sstate hit rate
- [x] Unchanged cache: push logs fingerprint skip for each type
- [ ] Missing zstd on runner: falls back to zip without failing the job hard

### DIFF
--- a/.github/scripts/s3-bitbake-cache.sh
+++ b/.github/scripts/s3-bitbake-cache.sh
@@ -206,12 +206,40 @@ push_all() {
   done
 }
 
+# Sum ContentLength of the .tar.zst archives pull would fetch (not the whole
+# bucket — leftover .zip / other keys must not trip the 50GB skip gate).
+# Prints a single integer byte count on stdout.
+active_archive_size_bytes() {
+  local s3_prefix="$1"
+  local cache_bucket="${s3_prefix#s3://}"
+  cache_bucket="${cache_bucket%%/*}"
+  local sizeInBytes=0
+  local cachetype key len
+
+  for cachetype in "${CACHE_TYPES[@]}"; do
+    key="${cachetype}.tar.zst"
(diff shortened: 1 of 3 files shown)


.github/scripts/s3-bitbake-cache.sh
```diff
@@ -0,0 +1,353 @@
...
+
+  if command -v zip >/dev/null 2>&1 && command -v unzip >/dev/null 2>&1; then
+    ARCHIVE_FORMAT=zip
+    log "WARNING: zstd unavailable; falling back to zip for cache archives"
+    return 0
+  fi
+
+  log "ERROR: neither zstd nor zip/unzip is available" >&2
+  return 1
+}
```
y3rsh: I do not mind a simple check here but I don't like a catch all installer.  I think we should understand if our ephemeral custom runners will or will not have zstd and reconcile them to have it where we control that reconciliation.  From a quick check it looks like almost all standard github runners do have it, just FYI.
This check can then simply be print the version or fail if not present.
I also think we don't need any backwards compatibility.  When this merges I expect no cache hit and new caches built and pushed according to this new pattern.

We want the code in this file as simple and straightforward as possible to make it clear to all maintainers exactly how we build, bust, write, and use the caches for this job.

sfoster1 (approved):
looks good to me!

PR #350, opened 2026-07-13
ci: STEP 3: fetch oe-core, monorepo, and firmware in parallel

neo-jesse:
## Summary
- Replace serial `actions/checkout` steps for oe-core / opentrons / ot3-firmware with one parallel `git clone` step.
- Wall time for source fetch is bound by the slowest repo (usually monorepo) instead of the sum.

## Test plan
- [ ] Stage-dev build checks out the decided refs correctly in `./oe-core`, `./opentrons`, `./ot3-firmware`
- [ ] Submodule `update.sh` still succeeds after parallel clones
- [ ] Compare checkout wall time to a prior serial run

### DIFF
--- a/.github/workflows/build-ot3-actions.yml
+++ b/.github/workflows/build-ot3-actions.yml
@@ -273,32 +273,52 @@ jobs:
           submodules: false
           fetch-depth: 0
           path: ./oe-core-for-workflow
-      - name: Fetch oe-core source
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
ok, this looks good to me. i'm a little nervous because the git version stuff we do is sensitive to exact git configurations, but let's try it and see how it works.

PR #351, opened 2026-07-13
ci: STEP 4: skip git S3 archive; persist pnpm/electron/pip caches

neo-jesse:
## Summary
- Add `S3_BITBAKE_CACHE_TYPES` allowlist to the BitBake S3 cache helper.
- CI sets `downloads,sstate,pnpm,electron,pip,pip-buildenv` (skips the ~13GB `git` archive).
- Create local cache dirs for the new types and document the allowlist.
- `active-size-bytes` respects the same allowlist (so leftover `git.tar.zst` does not inflate the 50GB gate).

## Notes
Retargeted onto `main` after the tar.zst PR (#349) merged. Rebased to resolve conflicts with the current helper (require-zstd / active-size-bytes).

## Test plan
- [ ] Pull logs `Using cache types: downloads sstate pnpm electron pip pip-buildenv`
- [ ] Pull wall time drops vs runs that fetched `git.tar.zst`
- [ ] After a build that populates pnpm/pip, push uploads new archives (or skips when unchanged)

### DIFF
--- a/.github/scripts/s3-bitbake-cache.sh
+++ b/.github/scripts/s3-bitbake-cache.sh
@@ -2,9 +2,14 @@
 # BitBake S3 cache for oe-core CI.
 #
 # Local trees (under <cachedir>/):
-#   downloads/   BitBake DL_DIR
-#   sstate/      BitBake SSTATE_DIR
-#   git/         BitBake GITDIR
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
ok, sure

PR #352, opened 2026-07-13
feat(build): STEP 5: shared pip-buildenv for Opentrons app recipes

neo-jesse:
## Summary
- Honor `PIP_CACHE_DIR` in `opentrons_app_bundle.bbclass`.
- Share a flock-guarded host `pip-buildenv` across parallel server recipes when the cache dir is set.
- Use per-recipe private pip caches (hardlink-seeded) during `do_compile`, then merge back under flock to avoid pip lock contention.
- Keep `--no-binary :all:` by default via overridable `OPENTRONS_APP_BUNDLE_PIP_NO_BINARY`.
- Register new variables in `oelint-constants.json`; set `PIP_CACHE_DIR` in CI `local.conf`.

## Test plan
- [ ] oelint passes on `opentrons_app_bundle.bbclass`
- [ ] Stage-dev build: logs show shared buildenv reuse across server recipes
- [ ] Server `do_compile` wall is not worse than pre-change ~14.5m (ideally better on warm pip-buildenv)
- [ ] Default still uses `--no-binary :all:` unless overridden


sfoster1 (approved):
ok, let's see how it goes. i'll admit that i don't quite see the point of this if we're keeping `--no-build-isolation`, i wouldn't have thought the pip buildenv step was at all a limiting factor, but it's nice to make things faster when we can I suppose.

PR #354, opened 2026-07-16
layers/meta-opentrons: OT-Protocol user permissions for logs and protocols directory

CaseyBatten:
Partial coverage of EXEC-2826

This PR adds the `ot-protocol` user to the `adm` and `systemd-journal` user groups to enable logging. This is to be paired with a monorepo PR which adds permissions to the protocols directory used by protocol runs for the `ot-protocol` user.

This PR also adds a tempfiles approach to directory permission validation. On boot the `opentrons-robot-server-permissions.conf` file will ensure schema >=17 directories have the proper permission heirarchy.


CaseyBatten (commented):
Reverted change removing the file permissions tempfile behavior to experiment with directory sanitization in situations where file source originates outside of normal protocol upload, particularly when the `/protocols` directory doesnt exist yet.

sfoster1 (approved):
Looks good overall, but i don't quite see the point of having _both_ open world-permissions (LSB of the permission bits) _and_ group management. presumably the point of setting the group here would be to keep world permissions tight?

it also would be really nice to figure out a way to _not_ let people edit `/var/lib/opentrons-robot-server` from inside a protocol. that's where protocols are stored, and we want to secure them against messing around.

CaseyBatten:
> Looks good overall, but i don't quite see the point of having _both_ open world-permissions (LSB of the permission bits) _and_ group management. presumably the point of setting the group here would be to keep world permissions tight?
> 
> it also would be really nice to figure out a way to _not_ let people edit `/var/lib/opentrons-robot-server` from inside a protocol. that's where protocols are stored, and we want to secure them against messing around.

Permissions are now restricted by group with no user field, and since these are read and execute only (with the exception of sensor logs specifically) the `opentrons-robot-server` persistence directory should be safe.

PR #356, opened 2026-07-21
fix(ci): STEP 0: size S3 cache skip gate on active archive keys only

neo-jesse:
## Summary
- The pull-cache **>50GB skip** used `aws s3 ls --recursive --summarize` on the whole `S3_CACHE_ARN` bucket.
- Buckets that still hold both `.zip` (what main pulls) and leftover `.tar.zst` from experiments report ~61GB and skip cache entirely → multi-hour cold BitBake.
- Sum only the archives this job actually pulls (`downloads` / `sstate` / `git` + the active extension) via `head-object` ContentLength.

**Note on zip vs zstd:** main does **not** choose format based on whether `zstd` is installed — it always pulls/pushes `.zip`. This PR keeps that and sizes `.zip` only. When the tar.zst cache path (#349) lands, flip `cache_archive_ext` (and the extract step) so the gate stays aligned with pull.

## Test plan
- [ ] stage-dev Flex build with a mixed zip+zst bucket: log shows per-key sizes totaling ~27GB and **does** pull cache (no “clean build without cache” from the 61GB whole-bucket sum)
- [ ] Missing objects still count as 0 and pull retries/skips as before
- [ ] Confirm IAM: `s3:GetObject` / head-object on those keys (no longer depends on ListBucket for the size gate)

### DIFF
--- a/.github/workflows/build-ot3-actions.yml
+++ b/.github/workflows/build-ot3-actions.yml
@@ -405,24 +405,52 @@ jobs:
       - name: Pull S3 cache
         shell: bash
         run: |
-          # fetch cache if the size is less than 20GB, so we have enough space to build image.
-          sizeInBytes=`aws s3 ls s3://${S3_CACHE_ARN/arn:aws:s3:::/} --recursive --summarize | awk END'{print}' | grep -o [0-9].*`
-          sizeInGigabytes=$(($sizeInBytes/1024/1024/1024))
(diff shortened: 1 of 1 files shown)


y3rsh (approved):
Looked up the 
`^[0-9]+$` means: the whole string is one or more digits — nothing else.
Match → treat as a byte count and add it
No match → don’t add; log “missing”

Makes sense.  Might change the variable name `localzip` in one of the the other PRs

PR #357, opened 2026-07-21
meta-opentrons: include robot-server uds in audit-server service file

jbleon95:
Addresses part of EXEC-2779.

For the audit-server log period download, we need to get the robot's name and serial number to write to a robot identity file. To do so, we need to ping the robot-server's `/health` endpoint to get this information. Now on audit server boot we will look for one of `OT_AUDIT_SERVER_robot_server_url` or `OT_AUDIT_SERVER_robot_server_uds` to establish connection with the robot server. This PR adds the latter, allowing us to talk with the robot-server when the audit server is running on a Flex.


SyntaxColoring (approved):
I would use robot-server’s Unix domain socket for this instead of its public-facing URL, just for a little more efficiency and to take nginx out of the loop, but this LGTM either way if it’s been tested. Thanks!

sfoster1 (approved):
this will work, talking to the robot server this way will actually go through nginx, which is a bit annoying. you could instead tell it to talk to the robot server via `/run/aiohttp.sock` (we should change this at some point lol) which is how nginx talks to the robot server: https://github.com/Opentrons/oe-core/blob/main/layers/meta-opentrons/recipes-robot/opentrons-robot-server/files/opentrons-robot-server.service#L19

SyntaxColoring (approved):
🧦

PR #358, opened 2026-07-22
fix(ci): treat ECR BuildKit failures as cache misses

neo-jesse:
## Summary
- Drop placeholder string checks for `OT3_OE_DOCKER_ECR_REPOSITORY_*`.
- Any ECR/BuildKit cache failure (bad URI like `na`/`missing`, auth, push denied) is treated as a **cache miss**: warn and fall back to plain `docker build`.
- ECR login is `continue-on-error` so a failed login still reaches the build step.

## Test plan
- [ ] stage-prod with `OT3_OE_DOCKER_ECR_REPOSITORY_PROD=na`: warning + uncached docker build; job continues
- [ ] stage-dev with a real URI: BuildKit ECR cache still used on success
- [x] Optional: clear or set a real prod URI when ready for prod cache


sfoster1 (changes requested):
definitely don't do this this way. make an ecr interaction that fails just cause a cache miss instead. makes it more robust (what if someone puts "missing" in there?)

y3rsh:
also see the changes in #359

sfoster1 (approved):
nice, looks good once merge conflicts are fixed

neo-jesse:
Closing: main already covers this via #359 (placeholder ECR var denylist + BuildKit cache fallback to plain `docker build`). No need to land the overlapping changes here.

PR #360, opened 2026-07-24
fix(networkmanager): recover mlan0 when NM is connected without carrier

vegano1:
## Summary

Sometimes Flex Wi-Fi gets stuck where NetworkManager still reports `mlan0` **connected** (with an IP) but the radio has **no carrier** and is not associated. The UI/API say Wi-Fi is up, but the Wi-Fi IP is unreachable. Ethernet can still work.

This usually shows up after the link has been up for a while (roam/rekey/driver hiccup). `wpa_supplicant` may log `Reject scan trigger since one is already pending`, and NetworkManager never emits a useful `down` event, so the existing crash-only `bounce-mlan0` path never runs.

Closes: [RQA-5751](https://opentrons.atlassian.net/browse/RQA-5751)

### Changelog

- Add a `check` mode: if NM says `mlan0:connected` and carrier is 0, soft-reconnect with `nmcli connection down/up`.
- Run that check on a oneshot timer every 60s (`bounce-mlan0-check.timer`).

## Test plan
- [ ] Build/install image with this change
- [ ] Confirm timer is active: `systemctl status bounce-mlan0-check.timer`
- [ ] Healthy Wi-Fi: check is a no-op (`bounce-mlan0.sh check` / timer logs quiet)
- [ ] On a desynced robot (NM connected + `NO-CARRIER`): timer or manual `bounce-mlan0.sh check` restores association and Wi-Fi IP reachability
- [ ] Ethernet stays up during soft reconnect

[RQA-5751]: https://opentrons.atlassian.net/browse/RQA-5751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/layers/meta-opentrons/recipes-connectivity/networkmanager/files/mlan0-desync-check.service
+++ b/layers/meta-opentrons/recipes-connectivity/networkmanager/files/mlan0-desync-check.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=Opentrons mlan0 Wi-Fi desync check
+Documentation=file:///usr/sbin/mlan0-desync-check.sh
+After=NetworkManager.service
+Wants=NetworkManager.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/sbin/mlan0-desync-check.sh
+TimeoutStartSec=90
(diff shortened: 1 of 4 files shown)


CaseyBatten (approved):
You're my hero! These seem like sensible checks.

sfoster1 (changes requested):
the general idea looks good but let's do some or all of the following:
- Have this and the rest of the bounce logic take the same actions (e.g. either `nmcli conn down/up`  or `sdhci-esdhc-imx/unbind / bind` but not both)
- Have this and the rest of the bounce logic a little more tightly integrated (dedicated argument handling block that calls one of the two outcomes)
- just a separate script for this method

having these two things that go to the same purpose but take very different methods of invocation and execution is weird

sfoster1 (approved):
nice, looks good to me if it works!

PR #361, opened 2026-07-28
opentrons-robot-server: add version dropfile to hw

sfoster1:
The hardware api service needs the same version dropfile as the robot server, because that service also import opentrons.config which is what uses the presence of that version directive to know it's a flex.


SyntaxColoring (approved):

CaseyBatten (approved):
Great catch!

PR #364, opened 2026-07-30
opentrons-robot-server: alter hwapi startup check

sfoster1:
This now checks for the presence and truthiness of a new feature flag that disables hardware subprocesses. This is annoying compared to swapping the default value of the old feature flag, but since migrations happen in the robot server and the robot server doesn't start until after the hardware server, we can't rely on migrations here. This way, the process will start unless the migration has already run and a user has explicitly flipped the flag. the new flag is `disableHardwareSubprocess`.

there's a bunch of extra shell stuff in there so that the condition is true if the feature_flags file doesn't exist (like on first boot) or it's mangled.


CaseyBatten (approved):
Yep this covers the logical loop on migrations well! Nice catch and thanks Seth!

PR #366, opened 2026-08-04
layers/meta-opentrons: remote access checks

sfoster1:
Add a systemd unit that will succeed if remote access is allowed. That means
- Auth server is up and CRS is disabled
- Auth server is up and /etc/.allow-remote-access exists

That last file is on the read-only file system in a directory owned by root; you can only make it through a serial cable and only by remounting the rootfs, and then regardless of what you do it'll be blown away by a software update before the system is accessible again.

We can then set `After=` and `Requires=` dependencies on it in any unit that we deem requires remote access, and that unit won't start if this unit fails. For Jupyter, that means until reboot (or until the jupyter-notebook service is retriggered, anyway) but if you add or remove the file, or add or remove CRS, that means that ssh and devtools will immediately start/stop accepting new connections. Any active connections will not be terminated.

Closes EXEC-2549
Closes EXEC-2758
Closes EXEC-2550

## to come out of draft
- [x] copy these files on a robot to test
- [x] get a successful build
- [x] get that build on a robot to test

### DIFF
--- a/layers/meta-opentrons/recipes-robot/opentrons-auth-server/files/check_remote_access_allowed
+++ b/layers/meta-opentrons/recipes-robot/opentrons-auth-server/files/check_remote_access_allowed
@@ -1,0 +1,48 @@
+#!/usr/bin/python3
+
+import asyncio
+import sys
+import os
+import traceback
+import json
+
+import logging
+
+logging.basicConfig(level=logging.ERROR)
+
+
+async def _do_check():
+    import aiohttp
+
+    uds = os.environ["CHECK_REMOTE_ACCESS_auth_server_uds"]
+    connector = aiohttp.UnixConnector(path=uds)
+    async with aiohttp.ClientSession(
+        connector=connector,
+        base_url="http://localhost",
+    ) as session:
+        async with session.get("auth/settings/accessControlEnabled") as response:
+            response.raise_for_status()
+            response_bytes = await response.read()
+
+    parsed_response = json.loads(response_bytes)
+    if parsed_response["data"]["accessControlEnabled"] is False:
+        print("Allowing remote access: CRS is disabled")
+        return 0
+    path = os.environ["CHECK_REMOTE_ACCESS_override_file"]
+    if os.path.exists(path):
+        print("Allowing remote access: override file exists")
+        return 0
+    print("Disallowing remote access: CRS is enabled and no override exists")
+    return 1
+
+
+async def check():
+    try:
+        return await _do_check()
+    except Exception:
+        print("Disallowing remote access: error in check process")
+        traceback.print_exc()
+        return 2
+
+
+sys.exit(asyncio.run(check()))
--- a/layers/meta-opentrons/recipes-robot/opentrons-auth-server/files/check_remote_access_allowed
+++ b/layers/meta-opentrons/recipes-robot/opentrons-auth-server/files/check_remote_access_allowed
@@ -14,7 +14,7 @@ if "/opt/opentrons-auth-server" not in sys.path:
     sys.path.append("/opt/opentrons-auth-server")
 
 
-async def _do_check():
+async def _http_check() -> bool:
     import aiohttp
 
     connector = aiohttp.UnixConnector(path="/run/opentrons-auth-server.sock")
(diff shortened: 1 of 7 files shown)


layers/meta-opentrons/recipes-robot/opentrons-auth-server/files/check_remote_access_allowed
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/python3
```
caila-marashaj, line 1: did u mean to get rid of this comment + does this file need a file extension

layers/meta-opentrons/recipes-robot/opentrons-auth-server/files/opentrons-remote-access-allowed.service
```diff
@@ -0,0 +1,13 @@
...
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/bin/opentrons_check_remote_access_allowed
+
+
+[Install]
+RequiredBy=dropbear@.service
+RequiredBy=jupyter.service
```
vegano1: Should be `jupyter-notebook.service` instead of `jupyter.service`

layers/meta-opentrons/recipes-robot/opentrons-auth-server/files/opentrons-remote-access-allowed.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Opentrons remote-access-allowed trigger service
+After=opentrons-auth-server.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/bin/opentrons_check_remote_access_allowed
```
vegano1, line 8: Do we want this to pass an explicit TIMEOUT?

Without it, the checkers `timeout_or` then returns `None`, and `_do_check` retries forever, in combination with `After=opentrons-auth-server.service` for ordering only. So if the auth-server is broken or can't start for whatever reason, the `opentrons-remote-access-allowed.service‎` will stay "activating" indefinitely, effectively blocking all dependent services from starting.

Unless that's the intended behavior?

layers/meta-opentrons/recipes-robot/opentrons-auth-server/files/check_remote_access_allowed
```diff
@@ -0,0 +1,106 @@
...
+import os
+import traceback
+import json
+
+import logging
+
+logging.basicConfig(level=logging.ERROR)
+
+if "/opt/opentrons-auth-server" not in sys.path:
+    sys.path.append("/opt/opentrons-auth-server")
```
vegano1, line 14: What is this doing? It does not look like we are importing any auth packages from this script. I see that you are importing `aiohttp` which is not part of the deps for auth-server, since it uses uvicorn/fastapi, and `opentrons-auth-server.bb` does not RDEPENDS on python3-aiohttp.

layers/meta-opentrons/recipes-robot/opentrons-auth-server/files/check_remote_access_allowed
vegano1: This is not a comment; it tells the OS what program to interpret the rest of this script with

vegano1 (commented):
had some questions

layers/meta-opentrons/recipes-robot/opentrons-auth-server/files/opentrons-remote-access-allowed.service
sfoster1: that is the intended behavior, yes. if the auth server isn't up, we don't know if CRS is on or not, and shouldn't allow remote access. Note though that if the magic file exists, we don't check the server at all

layers/meta-opentrons/recipes-robot/opentrons-auth-server/files/check_remote_access_allowed
sfoster1: `aiohttp` is part of the deps for auth-server via server-utils for clients to other servers. it's not installed in `/usr/lib/python3.12/site-packages` though, so we have to get it from one of the servers, and this is associated with auth, so i picked that one.

vegano1: gotcha, I forgot about server-utils!

vegano1 (approved):
Looking good to me !

PR #368, opened 2026-08-07
layers/meta-opentrons: add a way to disable crs

sfoster1:
It's a script! opentrons_disable_crs. You need the same password you do to enable it, and it just blows away the auth server state dir and restarts.

Closes EXEC-2176


mjhuff (approved):
Very cool-0000

PR #369, opened 2026-08-10
fix(networkmanager): enable mlan0-desync-check.timer at boot

vegano1:
## Summary
- Enable `mlan0-desync-check.timer` at boot via a static `timers.target.wants` link (same pattern as `opentrons-init-systemconnections`).
- Package the wants symlinks so they ship in the image.

## Why
The timer unit was installed but stayed disabled: package postinst only enables stock NetworkManager units, and the distro preset is `disable *`. Desync recovery never ran unless started manually.

## Test plan
- [ ] Build image and confirm `/usr/lib/systemd/system/timers.target.wants/mlan0-desync-check.timer` exists
- [ ] After boot, `systemctl is-active mlan0-desync-check.timer` is active
- [ ] Timer fires (journal shows desync check runs after ~90s, then every ~60s)


sfoster1 (approved):
looks good to me if it builds and runs!

PR #370, opened 2026-08-10
fix: Fix issues when running a first-flash on a robot

ryanthecoder:
There was two issues that were messing up first-flashing a robot on 9.0+.

First when we updated linux, uboot got an update but this newer version changed the way that the u-boot-initial-env-sd file inside the full image tarball got created. It was missing one key value (`bootcmd=run distro_bootcmd`) and there was several memory address changes that were leading to the u boot script not working correctly. 
For example in versions up to and including 8.8 
`kernel_addr_r=0x48200000`
and after 9
`kernel_addr_r=0x42000000`
Consequently it would prevent the part of the script that detected and the SOM revision and correctly changed the can clocking rate. The system would boot but it would be the wrong clock rate and no can messages could be sent. 

The other issue was that the tar command was not correctly preserving the root:root ownership of files resulting in userfs partition being owned by westen:westen. When the system would boot up systemd could not create tempfiles in /var since root:root did not own the directory and this would block many of the systemd services such as systemd-networkd from starting. 
Updating the tar command in the script to force the root:root ownership fixed the issue.


sfoster1 (approved):
ok, we can do it and let tmpfiles override it

PR #373, opened 2026-08-12
layers/meta-opentrons: Ensure pyro systemd services restart as group

CaseyBatten:
Covers: EXEC-2677, RQA-5790, RQA-5789

This PR updates the `opentrons-pyro-nameserver`, `opentrons-robot-server` and `opentrons-hardware-api` services to ensure that if one of them restarts all of them restart. This should solve our pain points of something like the robot server losing it's hardware api resource.

In the future we may want to make these services intelligently re-find eachother, instead of restarting when one goes down.


sfoster1 (approved):
ooh i didn't know about this, that's sweet

